PR #3, opened 2016-08-17
Lifecycle event name changed

kpotehin:
Several lifecycle event names for deploy plugin [have been renamed](https://github.com/serverless/serverless/issues/1577#issuecomment-238226700) in the latest version of Serverless. Here's the fix.

### DIFF
--- a/index.js
+++ b/index.js
@@ -73,7 +73,7 @@ class ServerlessWebpack {
     };
 
     this.hooks = {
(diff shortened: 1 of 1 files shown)


thenikso:
Thank you I spotted this a few minutes ago as well. Version `serverless-webpack@1.0-beta.2` should already be on npm with the fix!

Issue #4, opened 2016-08-18
Multiple Entry?


mikecann:
Hi,

Awesome work.

Currently the webpack.config.js entry only supports a single entry point "handler.js". How would one go about having multiple functions in different files? Perhaps you could update the example to show how you would have multiple functions? 

Thanks so much again,
Mike

thenikso:
I got this working in #7, it will be part of beta 3

PR #7, opened 2016-08-18
Multiple webpack entrypoint support

thenikso:
Starting from #4 this PR adds support for multiple entrypoints as well as fixing a problem with handler loading. 

Before this pr it was assumed that the function id was the same as the handler so:

```
functions:
  hello:
    handler: handler.hello
```

Worked, while:

```
functions:
  hello:
    handler: handler.something
```

would not. This should now be fixed.

TODO:

- [x] support multiple endpoints
- [ ] add tests

### DIFF
--- a/examples/multiple-entries/event.json
+++ b/examples/multiple-entries/event.json
@@ -1,0 +1,5 @@
+{
+  "key3": "value3",
+  "key2": "value2",
+  "key1": "value1"
+}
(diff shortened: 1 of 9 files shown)


andymac4182:
What do you think of loading the entry points from the `serverless.yml` file instead of having to define them in the `webpack.config.js` file?

thenikso:
I'm not very positive about it. I'd keep the concerns separate for now. it might become reasonable later on

PR #9, opened 2016-08-21
allow custom output filename

mikestaub:
Currently webpack configs that specify a custom output filename will break this plugin. This fixes that.

### DIFF
--- a/lib/run.js
+++ b/lib/run.js
@@ -8,7 +8,7 @@ const utils = require('./utils');
 module.exports = {
   loadHandler(stats, functionId, purge) {
     const handler = this.serverless.service.functions[functionId].handler.split('.');
-    const moduleFileName = `${handler[0]}.js`;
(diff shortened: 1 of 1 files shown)


thenikso:
however if the `handler` first part does not correspond to the bundle filename, AWS lambda itself would not find the function handler right?

In that case having the simulator fail would be the correct behavior I'd say. Probably a better error report would make sense. 

can you give an example of what situation this fix would solve?

mikestaub:
Good point, I didn't think of that. In that case, yes a better error would be useful.

I was just testing this plugin out and I was trying to use my old webpack config which had a custom output filename.

But what you say makes sense, so you can close this PR.

thenikso:
Closing it then. However if you could come up with a better error about that case it would be great. Maybe we should also update the readme to make that handle naming thing more clear.

PR #17, opened 2016-08-27
Properly send CORS headers for options request

janicduplessis:
The old options handler did not send CORS headers and this caused requests to fail when using browser `fetch`. This adds a separate options handler for each function and adds CORS headers if needed.

### DIFF
--- a/lib/serve.js
+++ b/lib/serve.js
@@ -39,23 +39,18 @@ module.exports = {
 
     app.use(bodyParser.json({ limit: '5mb' }));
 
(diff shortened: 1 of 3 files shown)


thenikso:
nice catch, basically all the CORS work was for nothing is the OPTIONS req was not returning the headers xD 

merging and adding to the next milestone.

thanks!!

janicduplessis:
@thenikso Could you publish an update with this fix on npm soon?

thenikso:
Yes I'll prepare the next version this weekend

PR #24, opened 2016-09-02
Update readme.md file

StefH:


thenikso:
actually `webpack.config.js` is correct as webpack uses a js file. the example itself is incomplete in that case. This feature will be modified in the next future and the readme as well. Closing this

thenikso:
woops that was the contrary :D good catch.

Issue #21, opened 2016-08-31
webpackIncludeModules ignores version in package.json


kennu:
When using webpackIncludeModules to install an external node module, the version specified in the Serverless project's package.json is ignored. Instead, the directive will just install the latest version published on npm, which might be completely different.

PR #26, opened 2016-09-07
support external-module-auto-packaging

cyyuen:
### external-modules-auto-packaging 

external-modules-auto-packaging feature allow user to defines `externals modules` in webpack and the plugin will parse the compilation stat, find out all external packages, include the excluded modules in node_modules within artifact if it is stated as production dependency in `package.json`. 

To enable this feature, simply defines `externals` in webpack config file and set webpackIncludeModules custom key to be true in serverless.yml. eg.

```js
// webpack.config.js
var nodeExternals = require('webpack-node-externals')

modules.export = {
  // we use webpack-node-externals to excludes all node deps. You can manually set
  // the externals too.
  externals: [nodeExternals()],
}
```

```yaml
# serverless.yml
custom:
  webpackIncludeModules: true # enable auto-packing
```

By default, the plugin will use the `package.json` file in working directory, If you want to
use a different package conf, set `packagePath` to your custom package.json. eg:

```yaml
# serverless.yml
custom:
  webpackIncludeModules:
    packagePath: '../package.json' # relative path to custom package.json file.
```
> Noted that only relative path is supported now.

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,23 +33,39 @@ generated to write bundles in the `.webpack` directory.
 By default, the plugin will try to bundle all dependencies. However, you don't
 want to include all modules in some cases such as selectively import, excluding
 builtin package (aws-sdk) and handling webpack-incompatible modules. In this case,
-you add all the modules, you want to exclude from bundled files, into `externals` field
-of your `webpack.config.json` and add those, you want to include in final distribution,
-into `serverless.yml`:
+you can enable external-module-auto-packaging feature by setting `webpackIncludeModules`
+custom key to be true and defines your webpack `externals` in `webpack.config.json`.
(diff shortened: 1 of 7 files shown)


cyyuen:
hey @thenikso . I have created the PR according to what we have discussed in #18 . Wanna have your feedback. 😄

thenikso:
I'll be checking this out asap. thanks @cyyuen! 

We should also think how to address #21

cyyuen:
@thenikso this new approach will also take the version into account. So #21 should be solved in this way.

thenikso:
I like it 👍 @cyyuen could you please fix the conflicts?

Also if you could also try it all with Serverless rc1 maybe you catch something I'm missing (also in the rest of the plugin). see #28

thanks!

PR #30, opened 2016-09-14
Add support for -d option (to generate source-maps)

StefH:
- Add support for -d option (to generate source-maps)
- Fix tests on Windows 7/10 (Note that 2 tests still fail ?)


thenikso:
I'd lave that to the webpack configuration itself. I don't really see a value added in having an option. What's your use case?

StefH:
I want to use your project as:

`"build:webpack": "rimraf ./dist && serverless webpack --out dist -d"`

Which will also build the source-maps.

hassankhan:
I tend to agree with @thenikso here, this is something I'd prefer my configuration to handle.

HyperBrain:
The webpack config now has access to the whole Serverless context and options. You can use these now to enable sourcemap support dynamically in your config. You even can react on any (new) options you might introduce in your environment by using the exported `lib.options` object.

```js
// webpack.conf.js

// This will check for a non standard CLI option --debug
const isDebug = slsw.lib.options.debug;

// Or you can make it dependent on the currently deployed stage
const isDebug = slsw.lib.options.stage === 'dev';

// Or even anything available in the current Serverless execution context
const isDebug = slsw.lib.serverless.XXXXXXXXX;
```

Closing this PR now as it is now possible to write fully dynamic webpack configurations.

PR #40, opened 2016-09-28
fixed issue causing cleanup function to fail due to .serverless folde…

id0Sch:
…r already exists

from now, move will only move the output of the build instead of the entire folder which is safer


id0Sch:
didn't see https://github.com/elastic-coders/serverless-webpack/pull/36 so i closed the PR

PR #45, opened 2016-10-11
Accept string format http events

### LINKED ISSUE #44: Shorthand http event syntax throws "Cannot read property 'toLowerCase' of undefined"
`serverless-webpack` `1.0.0-rc.2` at [this line](https://github.com/elastic-coders/serverless-webpack/blob/v1.0.0-rc.2/lib/serve.js#L44) assumes a serverless.yml event format like this:

``` yml
functions:
  myfunction:
    handler: handlers.myfunction
    events:
      - http:
          method: POST       # method and path are
          path: myfunction   # defined as separate keys
```

However using the http event string shorthand...

``` yml
functions:
  myfunction:
    handler: handlers.myfunction
    events:
      - http: POST myfunction

```

...throws the error...

```
Cannot read property 'toLowerCase' of undefined
... serverless-webpack/lib/serve.js:45:40 ...
```

marcuswhybrow:
Shims in an httpEvent object of the correct shape if the original httpEvent is a string.

### DIFF
--- a/lib/serve.js
+++ b/lib/serve.js
@@ -44,6 +44,15 @@ module.exports = {
 
     for (let funcConf of funcConfs) {
       for (let httpEvent of funcConf.events) {
+        if (typeof httpEvent === 'string') {
+          let split = httpEvent.split(' ');
+          if (split.length >= 2) {
+            httpEvent = {
+              method: split[0],
+              path: split[1]
+            }
+          }
+        }
         const method = httpEvent.method.toLowerCase();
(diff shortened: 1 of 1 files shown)


thenikso:
love it! 

I just saw that I broke a test in master. I'll fix that. 
Could you also add a test for this new feature?

marcuswhybrow:
I'm not familiar with JS test libraries, will try to explore that when I get some time.

HyperBrain:

Issue #19, opened 2016-08-29
support multiple webpack config files


mikestaub:
Currently this plugin only supports one webpack config file that is used during local development and during ```serverless deploy```.

I would like to have two different webpack configurations, one for each of those scenarios.

These could be specified in serverless.yml like so:

```
custom:
  webpack: 
    dev: ./webpack.dev.config.js
    deploy: ./webpack.deploy.config.js
```

The default behavior would still apply: if either was not specified, 'webpack.config.js' will be used.

What do you think?

mikestaub:
Currently, I'm setting an environment variable locally and checking for that in my webpack.config.js file. I used that to determine which config to use. Perhaps this is actually the correct way to solve this, I'm not sure.

andymac4182:
@mikestaub I am currently doing this at the moment for this in my project.
`serverless.yml`
```
custom:
  webpack: ${webpackConfigFile}
```

`serverless.env.yml`
```
vars:
  webpackConfigFile: webpack.local.config.js
stages:
  prod:
    vars: 
      webpackConfigFile: webpack.octopus.config.js
```

hope this helps

mikestaub:
@andymac4182 Thanks, that's a much cleaner solution!

PR #47, opened 2016-10-13
Added hooks to transpile with serverless-webpack before starting serverless-offline.

### LINKED ISSUE #29: Work with serverless-offline plugin
The `serverless-offline` plugin has a large support for handling most of the emulation of API Gateway. It would be great if I could use this plugin with that one to get a great offline experience and the great processing experience I am having with this plugin. 

Not sure if this is something you can work with the author of that plugin on.

gertjvr:
This will allow us to combine serverless-webpack and serverless-offline to get the best of both plugins. 

This will resolve #29 when the hooks are include in the serverless-offline plugin in a seperate pull request [#19](https://github.com/dherault/serverless-offline/pull/119).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,5 @@ node_modules
 dist
 .webpack
 coverage
+
+.idea
--- a/index.js
+++ b/index.js
@@ -117,6 +117,14 @@ class ServerlessWebpack {
       'webpack:serve:serve': () => BbPromise.bind(this)
         .then(this.validate)
         .then(this.serve),
+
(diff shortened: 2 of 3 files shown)


thenikso:
amazing! could you add an example to show a setup with both plugins?

also do you think that `serverless webpack serve` could be dropped at this point?

gertjvr:
Think so, but depends how many users depends on serve at the moment.

gertjvr:
Need to change the target branch for this pr.

PR #48, opened 2016-10-16
fix for deploying individual packages.

gertjvr:
Serverless added the ability to deploy functions [individually](https://github.com/serverless/serverless/blob/9ebf296762e502d42eba10108cea4bda94fa0c09/docs/01-guide/10-packaging.md#packaging-functions-separately) when using `serverless-webpack` it doesn't change the paths of the individual artifacts similar to how it change the this.serverless.package.artifact.

This solutions is very brittle as it copies how individual packages was implemented.

PR #49, opened 2016-10-18
Add before start hook

gertjvr:
This is a new PR to replace #47

### DIFF
--- a/lib/cleanup.js
+++ b/lib/cleanup.js
@@ -18,6 +18,19 @@ module.exports = {
             if (err) {
               reject(err);
             } else {
+              if (this.serverless.service.package.individually) {
+                const functionNames = this.serverless.service.getAllFunctions();
+                functionNames.forEach(name => {
+                  this.serverless.service.functions[name].artifact = path.join(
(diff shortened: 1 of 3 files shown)


gertjvr:
In response to https://github.com/elastic-coders/serverless-webpack/pull/47#issuecomment-253528703 I have created an [example](https://github.com/ticketsolutionsptyltd/serverless-offline-example) of how I would use `serverless-offline` and `serverless-webpack`

gertjvr:
any update on this PR and when it will be merged?

lib/cleanup.js
```diff
@@ -18,6 +18,19 @@ module.exports = {
             if (err) {
               reject(err);
             } else {
+              if (this.serverless.service.package.individually) {
```
thenikso: could you update with master? this part should already be present from #48

gertjvr: Will have a look at it tomorrow.

thenikso: thanks! also #55 looks similar to this PR but it also has readme improvements. would you review that one as well please? 🍰

PR #55, opened 2016-11-07
Add compatibility with serverless-offline plugin

### LINKED ISSUE #29: Work with serverless-offline plugin

miltador:
This PR adds basic compatibility with [`serverless-offline`](https://github.com/dherault/serverless-offline) plugin.
It hooks the `serverless-offline` event, validates and compiles functions before running the plugin.
I also added a note to the README about it.

Closes #29

### DIFF
--- a/index.js
+++ b/index.js
@@ -117,6 +117,10 @@ class ServerlessWebpack {
       'webpack:serve:serve': () => BbPromise.bind(this)
         .then(this.validate)
         .then(this.serve),
+
+      'before:offline:start': () => BbPromise.bind(this)
+        .then(this.validate)
+        .then(this.compile),
     };
   }
 }
--- a/README.md
+++ b/README.md
@@ -190,3 +190,20 @@ plugin with Babel. To try it, from inside the example folder:
 
 - `npm install` to install dependencies
 - `serverless webpack run -f hello` to run the example function
+
+
+## Compatibility with other plugins
+
+### [`serverless-offline`](https://github.com/dherault/serverless-offline) 
+
+Add both plugins to your `serverless.yml` file:
+
+```yaml
+plugins:
+  - serverless-webpack
+  - serverless-offline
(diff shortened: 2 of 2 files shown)


miltador:
Sorry, haven't noticed #49 before. It basically does the same but I added the note about it in README. It's up to you which one to merge.

miltador:
I also think that `serverless webpack serve` is obsolete now, it doesnt handle request templates correctly, e.g. `/api/path/{id}` endpoints are not reachable.

utkuturunc:
any news?

thenikso:
#49 has another hook on `before:offline:start:init` is that redundant? I'm liking this one more.

Also would it make sense to remove all offline code from this plugin given the serverless-offline compatibility?

index.js
```diff
@@ -117,6 +117,10 @@ class ServerlessWebpack {
       'webpack:serve:serve': () => BbPromise.bind(this)
         .then(this.validate)
         .then(this.serve),
+
+      'before:offline:start': () => BbPromise.bind(this)
+        .then(this.validate)
+        .then(this.compile),
```
thenikso: just trowing this our there, would a "watch" here instead of a "compile" make sense?

rvaidya:
If I use this.watch rather than this.compile I get

const handler = this.serverless.service.functions[functionId].handler.split('.');
TypeError: Cannot read property 'handler' of undefined

this.watch is your own sls webpack watch function, not webpack watch

thenikso:
Implemented in #58

pietrasiak:
Anyone knows some example that has it implemented?

Issue #56, opened 2016-11-15
Implement a webpack watch command over all functions


miltador:
As I mentioned in #55, with the `serverless-offline` plugin the current `Simulate API Gateway locally` feature is pretty outdated and doesn't have so reach features as `serverless-offline` has. Nevertheless this plugin is great for bundling functions with webpack.
By working with this plugin and `serverless-offline` the only feature I miss is a watcher for changes in functions code with auto rebuilding.
This plugin has a `watch` command for separate function only. And a `serve` is overkill as it not only watches changes but hosts functions too.
It would be great to have a command for watching changes of the whole project code. In this case we probably want the plugin to run separately from `serverless-offline` so both can do their jobs.

utkuturunc:
I don't mind having both plugins serve. How did you get serverless-offline work with serverless-webpack's output? All I can do is have them both serve at the same time. One's output is not piped to other's input.

miltador:
@utkuturunc,
Watch #55, basically I added a hook into `serverless-webpack` so it makes a build before executing `serverless-offline`.  And I just run `serverless offline` and it builds my code and then serves it locally.

utkuturunc:
@miltador I am using a similar solution. I want to implement watching feature to this solution.

miltador:
@utkuturunc, I got it. I see it as 2 separate tasks running, one is webpack watcher and another one is `serverless offline`. `serverless-offline` plugin doesn't need a restart on code changes. Another approach I see is that webpack watcher will rebuild on code changes and then start `serverless-offline` plugin. I would prefer the second one as it is simpler for end users.

utkuturunc:
I think the first choice is better. I am not sure how to handle node_modules though.

thenikso:
I agree that we should have this plugin just do the build and leave the "offline" part to other plugins.

I still haven't grasped how to handle the hot-reloading though. I think it's a must have feature. I'll play around with #55. Any idea is welcome!

PR #58, opened 2016-11-17
Yet another implementation for compatibility with serverless-offline

rvaidya:
Replaces #55, #49, #47 

This runs webpack.watch rather than webpack.compile for developer sanity reasons.

### DIFF
--- a/index.js
+++ b/index.js
@@ -4,6 +4,7 @@ const BbPromise = require('bluebird');
 
 const validate = require('./lib/validate');
 const compile = require('./lib/compile');
+const wpwatch = require('./lib/wpwatch');
 const cleanup = require('./lib/cleanup');
 const run = require('./lib/run');
 const serve = require('./lib/serve');
@@ -18,6 +19,7 @@ class ServerlessWebpack {
       this,
       validate,
       compile,
+      wpwatch,
       cleanup,
       run,
       serve,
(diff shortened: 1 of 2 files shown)


rvaidya:
You may want to edit the implementation for semantics, I'm using the name "wpwatch" since "watch" is taken by the "webpack watch" function.

rvaidya:
One note for serverless-offline, right now the entry point is hardcoded to ./handler.js so Webpack's output file must also be set to ./handler.js

Referenced serverless-offline PR https://github.com/dherault/serverless-offline/issues/153

miltador:
@rvaidya, nice!
I think you can remove the `offline` part now and the old `watch` command according to discussing in #56 so that we'll leave the service emulation to the `serverless-offline` plugin.

mhodgson:
👍  LGTM

rvaidya:
Did not mean to include this second commit, planning on opening a separate issue for making webpack a peer dependency

thenikso:
hello everyone! sorry for my absence! in the coming days I'll work on addressing all the opened PRs and issues starting with this one. 

Is this preferred to #55 ?

mhodgson:
@thenikso yes I've been using this for a while and it works well.

bassbump:
I feel that webpack serve should be removed and ppl should rely on serverless-offline for local testing, just focus only the webpack part

PR #74, opened 2017-01-04
Allow custom cors headers to be supplied via serverless.yml

Whoaa512:


thenikso:
this looks all right, I also like the documentation. However wouldn't it be easier to use a `Access-Control-Allow-Headers: *` instead?

Whoaa512:
I agree that `'*'` is easier and will incorporate that change. 

Also I think we could get closer to the serverless configuration ([See](https://github.com/serverless/serverless/blob/7fc058df6393d5d63721025f0814e4b3f91f8ba7/lib/plugins/aws/deploy/compile/events/apiGateway/lib/validate.js#L251))

Will try to incorporate that as well and see what you think.

Whoaa512:
Closing in favor of using [`serverless-offline`](https://github.com/dherault/serverless-offline)

PR #75, opened 2017-01-05
Also parse url encoded stuff mate

Whoaa512:


lib/serve.js
```diff
@@ -41,6 +41,7 @@ module.exports = {
       limit: '5mb',
       type: (req) => /json/.test(req.headers['content-type']),
     }));
+    app.use(bodyParser.urlencoded())
```
thenikso: we use semicolons here :D

thenikso:
could you please elaborate on how this would be useful? 
Also would it be a transparent update for current code?

Whoaa512:
Actually upon consideration, I don't think this should be added. Since by default API Gateway doesn't give this to you, and I believe the goal of the offline gateway simulation should be as close as possible to the real thing.

Though to answer your question, yes the change would be transparent.

PR #81, opened 2017-01-10
Make stage path prefixing optional

Whoaa512:
By default you don't need path prefixing when you specify a stage


HyperBrain:
@Whoaa512  It would be great if you could comment in #135 if you use the serve functionality or not, so that we can decide how to proceed there.

Whoaa512:
Closing in favor of using [`serverless-offline`](https://github.com/dherault/serverless-offline)

PR #85, opened 2017-01-12
Set output headers when lambda proxy is used

foysavas:
API Gateway now recognizes `headers` as part of [Lambda Function output](http://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-set-up-simple-proxy.html#api-gateway-simple-proxy-for-lambda-output-format).

### DIFF
--- a/lib/serve.js
+++ b/lib/serve.js
@@ -122,6 +122,11 @@ module.exports = {
         }
 
         if (isLambdaProxyIntegration) {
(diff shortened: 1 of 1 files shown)


HyperBrain:
@foysavas  It would be great if you could comment in #135 if you use the serve functionality or not, so that we can decide how to proceed there.

HyperBrain:

PR #83, opened 2017-01-11
Make webpack a peer dependency so that webpack 2 can be used

rvaidya:
Sorry for the wait

PR #87, opened 2017-01-13
Make webpack a peer dependency

### LINKED ISSUE #62: Make webpack a peer dependency so users can opt to use webpack 2
As a peer dep, we can use whatever webpack version we want as long as it's above the minimum required version

https://github.com/rvaidya/serverless-webpack/tree/peerdep

utkuturunc:
Users can now use webpack 2 instead of 1.

Closes #62


utkuturunc:
Sorry, didn't notice #83

PR #90, opened 2017-01-24
Pack all dependencies defined in package.json

miltador:
This will solve various issues with 'implicit' dependencies that aren't directly used in your code but required by other modules.
The developer himself is responsible to moderate the `dependencies` list and update it accordingly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,8 +91,8 @@ custom:
   webpackIncludeModules: true # enable auto-packing of external modules
 ```
 
-All modules stated in `externals` will be excluded from bundled files. If an excluded module
-is stated as `dependencies` in `package.json`, it will be packed into the Serverless
+All modules stated in `externals` will be excluded from bundled files. Every module 
+stated as `dependencies` in `package.json` will be packed into the Serverless
 artifact under the `node_modules` directory.
 
 By default, the plugin will use the `package.json` file in working directory, If you want to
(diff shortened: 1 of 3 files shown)


HyperBrain:
I think this PR is counter-productive to the mechanisms and use cases webpack is used for.
When you use wepback, it optimizes the dependencies and removes unneeded ones - this is one of the basic concepts behind it. So only the webpack result states exactly which modules have to be packaged and which not.
Imo the user cannot know, what is needed and what is not. So I'd like to close this PR as won't do.

What I see as a possible improvement in the whole way, the webpack plugin handles the fetching of the npm modules is, that it could retrieve the physical node modules that it packages from the projects node_modules directory instead of using npm to download them on deploy, but that's out of scope for this PR and should be handled and discussed in a separate issue (it is not trivial and has some side-effects).

HyperBrain:
The packaging of implicit dependencies has been solved recently ( #186 ) and is available in the V3 pre-release version `3.0.0-rc.1`.
Now you can bundle node modules by whitelisting them in `node-externals` and all transient dependencies (implicit dependencies) will be bundled correctly. The needed modules are perfectly detected by webpack after its optimization phase.

HyperBrain:
Closing this PR now. The package handling has been partly rewritten and improved in V3. If anything should be improved further, we should open an issue for that (feature request), so that it can be phased in correctly.

miltador:
That's cool! Thanks. I already gave up on AWS but this sounds cool.

HyperBrain:
BTW: Sorry for the _very_ late responses 😔 . The project was kind of dead from February until we revived it in Juli, but now it's back and I'll try to maintain it as good as possible (including frequent releases, etc.)

Issue #60, opened 2016-11-22
Pack individual functions using the new packageFunction hook


m3rls:
Hi,

I've noticed that when trying to deploy individual functions (with `package -> individually: true` in serverless.yml) the change made in https://github.com/elastic-coders/serverless-webpack/pull/48 has fixed the exception (thank you!), but the function isn't being packed.  

In my use case, this leads to a ~19MB .zip being uploaded compared to a ~600kb .zip when I do a full deploy.

[A Pull Request](https://github.com/serverless/serverless/pull/2561) was merged and released along with the serverless 1.2.0 milestone yesterday that added a `before:deploy:function:packageFunction` hook, which I have (naively and probably incorrectly) added to `index.js` as follows:

```js
      'before:deploy:function:packageFunction': () => BbPromise.bind(this)
        .then(this.validate)
        .then(this.compile),
```

So now when I run `serverless deploy function --function <myFunction>` it packages it (as illustrated below):

### Without hook:
```
➜  api git:(master) ✗ serverless deploy function --function myFunction
Serverless: Deploying function: myFunction...
Serverless: Packaging function: myFunction...
Serverless: Uploading function: myFunction (19.9 MB)...
Serverless: Successfully deployed function: myFunction
```

### With hook:
```
➜  api git:(master) ✗ serverless deploy function --function myFunction
Serverless: Deploying function: myFunction...
Serverless: Bundling with Webpack...
Time: 2337ms
     Asset     Size  Chunks             Chunk Names
handler.js  1.92 MB       0  [emitted]  main
Serverless: Packaging function: myFunction...
Serverless: Uploading function: myFunction (606.73 KB)...
Serverless: Successfully deployed function: myFunction
```

Would it be possible to please have support for this hook added at some point?  I'm happy to make a pull request for it if the way I've added it is acceptable, but didn't want to just assume it was correct.

Thanks

activefx:
@m3rls would you be willing to share your webpack config? I'm having trouble packaging functions individually and would appreciate it

m3rls:
@activefx sure, it's pretty minimal:

```js
module.exports = {
  entry: './handler.js',
  target: 'node',
  module: {
    loaders: [
      { test: /\.json$/, loader: 'json' }, {
        test: /\.js$/,
        loaders: ['babel'],
        include: __dirname,
        exclude: /node_modules/,
      }
    ]
  }
};
```

[...]

PR #91, opened 2017-01-26
Add hook for deploying individual functions

johnf:
As discussed in #60

### DIFF
--- a/index.js
+++ b/index.js
@@ -101,7 +101,8 @@ class ServerlessWebpack {
 
       'before:deploy:function:packageFunction': () => BbPromise.bind(this)
         .then(this.validate)
-        .then(this.compile),
+        .then(this.compile)
+        .then(this.packExternalModules),
 
       'webpack:validate': () => BbPromise.bind(this)
         .then(this.validate),
--- a/index.js
+++ b/index.js
@@ -99,6 +99,10 @@ class ServerlessWebpack {
       'after:deploy:createDeploymentArtifacts': () => BbPromise.bind(this)
         .then(this.cleanup),
 
+      'before:deploy:function:packageFunction': () => BbPromise.bind(this)
+        .then(this.validate)
+        .then(this.compile),
+
       'webpack:validate': () => BbPromise.bind(this)
         .then(this.validate),
 


mhodgson:
I can verify that this works. Would love to get this merged.

mhodgson:
Actually @johnf, I think you need to add `.then(this.packExternalModules)` as well for this to work for functions with external modules.

johnf:
@mhodgson you're right. Have just fixed that

deathweaselx86:
This would be wonderful.

kiawins:
Would also love to see some docs go along with this PR to understand how it could be used.

johnf:
Hi Kia,

It's nothing specific to serverless-webpack, it is just enabling standard serverless functionality.

i.e. making sure that ```sls deploy function -f functionName``` works. This is a lot faster as it just re-publishes the new code to Lambda and doesn't run cloudformation.

Docs at https://serverless.com/framework/docs/providers/aws/cli-reference/deploy-function/

johnf:
@thenikso any chance of getting this merged for another rc before 1.0.0 release?

blakedietz:
Any chance we could pull this in?

SkippyZA:
Also would really like to get this in asap.

SkippyZA:
@thenikso is there any word on this PR? Would like to try avoid needing to fork this :/

ShaunEgan:
Would also really love this to be merged.

hassankhan:
Big +1, getting `Code storage exceeded` errors from AWS 😢

HyperBrain:
@hassankhan 
> Big +1, getting Code storage exceeded errors from AWS

I think that's worth an issue in Serverless. The reason most likely is, that SLS does set the DeletionPolicy of the deployed function versions to "Retain", which will never delete any deployed version. Then you'll run out of code storage sooner or later.

hassankhan:
I'm not quite sure where the bug is, @HyperBrain 😕 . Spent most of last night trying to repair things as there was an issue with bundling. I have 19 functions, which together add up to 40MB. 

With `package.individually = false`, it seems that the deployment ZIP gets uploaded 19 times to Lambda. When setting it to `true`, it creates 19 ZIPs, but unfortunately inside the ZIP all 19 functions are present.

I started getting those errors, and incorrectly assumed it was to do with individual Lambda size. After much tinkering with Webpack, I reduced individual function size but continued to receive the same error on various functions upon CloudFormation deploy.

I finally clued onto the fact that I'd hit the 75GB hard limit by going to the Web Console and clicking Lambda. However, the Console was reporting total Lambda size as just over 1GB. It was then when I remembered that all functions are kept unless pruned, but at this point CloudFormation was stuck in a `UPDATE_REQUIRES_ACTION` state.

After fixing that, I was finally able to remove the stack and redeploy. I'm still looking into why the packaging is weird - for now the only fix I have is to use [serverless-prune-plugin](https://github.com/claygregory/serverless-prune-plugin) till I figure it out.

HyperBrain:
@johnf @mhodgson @blakedietz @SkippyZA @ShaunEgan @deathweaselx86 @hassankhan 
According to the comment in (https://github.com/elastic-coders/serverless-webpack/pull/130#issuecomment-312349755) this PR is superseded - as it seems that the ZIP file is not copied to the sls temp dir correctly. Can you please check if #130 works for you? I'd like to merge that then in favor of this one.

johnf:
I think I commented in one of the others that I also prefer #130. I didn't close this for some reason though!
Will do so now.

PR #82, opened 2017-01-11
Simulate Lambda Proxy Integration

ktonon:
http://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-create-api-as-simple-proxy-for-lambda.html#api-gateway-create-api-as-simple-proxy-for-lambda-build

PR #92, opened 2017-01-29
Add support for Catch-all path variables

nrako:
```yaml
functions:
  foo:
    handler: handler.foo
    events:
      - http:
          path: foo/{bar+}
          method: get
```
> **Catch-all Path Variables** – Instead of specifying individual paths and behaviors for groups of requests that fall within a common path (such as /store/), you can now specify a catch-all route that intercepts all requests to the path and routes them to the same function. For example a single greedy path (/store/{proxy+}) will intercept requests made to /store/list-products, /store/add-product, and /store/delete-product.

source: https://aws.amazon.com/blogs/aws/api-gateway-update-new-features-simplify-api-development/

PR #82 has some work around this but, from my test, it's implementation doesn't match what AWS has. e.g the name variable doesn't need to be called `{proxy+}`. And the named path variable must be available in `event.pathParameters`.

### DIFF
--- a/lib/serve.js
+++ b/lib/serve.js
@@ -49,7 +49,9 @@ module.exports = {
         if (this.options.stage) {
           endpoint = `/${this.options.stage}${endpoint}`;
         }
-        const path = endpoint.replace(/\{(.+?)\}/g, ':$1');
+        const path = endpoint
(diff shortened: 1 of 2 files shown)


HyperBrain:

PR #80, opened 2017-01-09
Support the `any` http method with the .all express handler

Whoaa512:
API Gateway now allows you to specify proxy routes like so:

```yaml
functions:
  hello:
    handler: handler.hello
    events:
      - http:
          path: ''
          method: any
      - http:
          path: '/{proxy+}'
          method: any

```

Sources:
- https://aws.amazon.com/blogs/aws/api-gateway-update-new-features-simplify-api-development/
- https://aws.amazon.com/blogs/compute/easier-integration-with-aws-lambda-and-amazon-api-gateway/

PR #93, opened 2017-01-29
Add support for ANY method

nrako:
```yaml
functions:
  foo:
    handler: handler.foo
    events:
      - http:
          path: '/foo'
          method: any
```

> **ANY Method** – Instead of specifying individual behaviors for each HTTP method (GET, POST, PUT, and so forth) you can now use the catch-all ANY method to define the same integration behavior for all requests.

source: https://aws.amazon.com/blogs/aws/api-gateway-update-new-features-simplify-api-development/

Supplement #80 with covering test. 
And...⚠️  second commit also prevent OPTIONS handlers to always be created, but instead only create handler when `cors: true` is set. From my test this would be more close to AWS, even if express has some default automatic OPTIONS handlers (see https://github.com/expressjs/express/issues/2246). [needs-discussion]?

### DIFF
--- a/lib/serve.js
+++ b/lib/serve.js
@@ -56,8 +56,10 @@ module.exports = {
           handler = this._handlerAddCors(handler);
           optionsHandler = this._handlerAddCors(optionsHandler);
         }
-        app.options(path, optionsHandler);
-        app[method](
+        if (method !== 'any') {
+          app.options(path, optionsHandler);
+        }
+        app[method === 'any' ? 'all' : method](
           path,
           handler
         );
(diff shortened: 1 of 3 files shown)


HyperBrain:
@nrako  It would be great if you could comment in #135 if you use the serve functionality or not, so that we can decide how to proceed there.

nrako:
@HyperBrain I haven't circled around `serverless-webpack` much lately. Looks like `serverless-offline` could be the way to go, but I can't really give a well-thought opinion on the serve functionality. In any case feel free to close the PR if necessary.

HyperBrain:
@nrako Thanks for coming back. A lot of other people in the community work with `serverless-offline` together with `serverless-webpack` and it shows out that the offline plugin (which is centrally built around offline emulation) does the job pretty well and better than serve.
We will concentrate on the packaging and let the emulation stuff for plugins which are solely made for that 😄 .

HyperBrain:

PR #97, opened 2017-02-09
Update webpack version from 1.3.x to 2.2.1

AKIRA-MIYAKE:
For complicated applications using type scripts, webpack 2 is more sophisticated and convenient.

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,7 +32,7 @@
     "express": "^4.14.0",
     "fs-extra": "^0.26.7",
     "npm-programmatic": "0.0.5",
-    "webpack": "^1.13.1"
+    "webpack": "^2.2.1"
   },
   "devDependencies": {
     "chai": "^3.5.0",


Gomah:
It's a breaking change, another PR https://github.com/elastic-coders/serverless-webpack/pull/83 was just merged about it.

domaslasauskas:
Good point Gomah. I jumped the gun by merging #83 as looking at it now it was a breaking change too, I'll be reverting/replacing it with a version range instead. I'm planing to go through the backlog by fixing issues/taking pull requests over upcoming weeks.

HyperBrain:
@domaslasauskas I'm planning to get out a stable update with the most needed fixes as soon as possible.
What is your opinion about the merged #83 . Although it might be a breaking change (as the including project has to specify a webpack version as peer dependency - we currently even use master with Webpack 3 without problems) I'm not really sure what side-effects or update problems might arise additionally.
Can you give a short statement about that?

hassankhan:
Not to disparage @AKIRA-MIYAKE's work, but is this PR relevant after #83? I'd be in favour of closing, since we can now specify what version we need ourselves.

HyperBrain:
Closing this, as the next release (v 2.0.0) adds Webpack as peer dependency and allows any Webpack version to be used.

PR #101, opened 2017-02-15
Load all environment vars before trying to require any files.

### LINKED ISSUE #99: Some, not all, environment vars are missing
Environment variables are not all available to functions. Only the variables defined globally and those for the first function will be available when running locally. Example:

`serverless.yml`:

```
provider:
  environment:
    A: 'A'
functions:
  func1:
    environment:
      F1: '1'
  func2:
    environment:
      F2: '2'
```

In this setup, the only variables available to the local environment are `A` and `F1`.
If I move func2 on top of func1, it is however `A` and `F2`.

bebbi:
Multiple functions may share a single file and may immediately depend on the environment.

Closes #99.

### DIFF
--- a/lib/run.js
+++ b/lib/run.js
@@ -16,8 +16,6 @@ module.exports = {
     if (purge) {
       utils.purgeCache(handlerFilePath);
     }
-    //set environment before requiring, as imported modules will be immediately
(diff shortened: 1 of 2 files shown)


HyperBrain:

PR #103, opened 2017-02-15
Rewrite relative local npm package.json dependencies to work when ins…

mattsills:
…talled in the dist directory

### DIFF
--- a/lib/packExternalModules.js
+++ b/lib/packExternalModules.js
@@ -5,7 +5,7 @@ const fs = require('fs');
 const path = require('path');
 const npm = require('npm-programmatic');
 
-function getProdModules(externalModules, packagePath) {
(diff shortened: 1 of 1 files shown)


jdbevan:
It would be nice if there was some kind of automated test to go with this, but the code looks good and I need this fix too, so +1 :)

hassankhan:
Is this still an issue? I've used relative paths for package dependencies in the past and don't remember having this problem.

HyperBrain:
@mattsills Could you please give an update (see @hassankhan 's question)? Please verify it with version 2.2.0.

HyperBrain:
I will close this PR now - there hasn't been any response for some weeks now. We should check if `file:` dependencies work in V3 and open a new issue if they do not. As the codebase has changed a completely new fix is appropriate then.

PR #104, opened 2017-02-19
Add instructions on how to use with serverless-offline

darthtrevino:
Adding a brief instructional section on how to use this plugin with serverless-offilne. I had to dig around a while to discover that you need the --location flag.

### DIFF
--- a/README.md
+++ b/README.md
@@ -118,6 +118,21 @@ The normal Serverless deploy procedure will automatically bundle with Webpack:
 - Install Serverless Webpack as above
 - Deploy with `serverless deploy`
 
+### Use with serverless-offline
+
+- Put the serverless-webpack plugin before the serverless-offline plugin in your configuration file.
+```yaml
+plugins:
+  - serverless-webpack
+  - serverless-offline
+```
+
+- Let serverless-offline know where the bundles are built
+```
+> serverless offline start --location .webpack 
+```
+
+
 ### Simulate API Gateway locally
 
 To start a local server that will act like the API Gateway use the following command.


README.md
```diff
@@ -118,6 +118,21 @@ The normal Serverless deploy procedure will automatically bundle with Webpack:
...
+- Put the serverless-webpack plugin before the serverless-offline plugin in your configuration file.
+```yaml
+plugins:
+  - serverless-webpack
+  - serverless-offline
+```
+
+- Let serverless-offline know where the bundles are built
+```
+> serverless offline start --location .webpack
```
HyperBrain: This command line will work as soon as #131 is merged.
The alternative is omitting `start`, so we may add the serverless command without start here too:
`serverless offline --location .webpack `

hassankhan: Good catch, @HyperBrain!

HyperBrain:
@hassankhan Just merged #132 which also updates the serverless offline section. We should compare the two and see how they fit together.

hassankhan:
Should we perhaps add a note mentioning if you use custom paths for your `webpack.config.js`, you should remember to update the `--location` flag accordingly?

README.md
```diff
@@ -118,6 +118,21 @@ The normal Serverless deploy procedure will automatically bundle with Webpack:
 - Install Serverless Webpack as above
 - Deploy with `serverless deploy`
 
+### Use with serverless-offline
```
hassankhan: Perhaps change to 'Usage'?

HyperBrain: I will change it accordingly and also add the `--location` hint (as long as I'm allowed to do that). Otherwise I open a separate PR with the same changes (as they are trivial) and close this one to get 2.0.0 out later.

HyperBrain: @hassankhan Just saw, that we have 2 sections for serverless-offline. This one and one under "Simulate API Gateway locally". I will combine the two into this one (as we plan to remove serve) and commit that. Then we can just remove the whole "Simulate ..." section together with the serve implementation later.

HyperBrain:
Restructured serverless-offline doc in #144 and integrated everything discussed or changed here. So this one is obsoleted and can be closed.

PR #105, opened 2017-02-19
Update the serverless-offline startup hook

darthtrevino:
I'm having trouble with the serverless-offline startup hooks in the latest version of serverless-offilne. It seems like you have to bind to the `offline:start:init` event in order to have things boot properly into the serverless-offline cycle.

### DIFF
--- a/index.js
+++ b/index.js
@@ -120,7 +120,7 @@ class ServerlessWebpack {
         .then(this.validate)
         .then(this.serve),
 
-      'before:offline:start': () => BbPromise.bind(this)
+      'before:offline:start:init': () => BbPromise.bind(this)
         .then(this.validate)
         .then(this.wpwatch),
     };


FrancescoA:
When will this get merged? Seems pretty important.

RuslanZavacky:
Hey guys, any updates? :) would be awesome to have this merged :)

pietrasiak:
Any update?

HyperBrain:
@darthtrevino Is this still an issue? We're using the master branch since quite a long time (which does not have the changes), but using `serverless-offline` works quite well.

HyperBrain:
I think @Whoaa512 just got it working too, without this PR.

Whoaa512:
As long as you have both plugins in your serverless.yml (with offline as the last plugin) and you run the command `sls offline` both plugins should work properly.

HyperBrain:
The implementation in #131 is more complete and also keeps the `start` hook too. I will close this one in favor of #131. Can you please test with that?

HyperBrain:
I tested #131 also for incompatibilities when using offline without the start argument. Works. So I'll close this one in favor of #131 as it is more complete.

PR #106, opened 2017-02-23
Update CORS to use allowed headers from yml

dolfelt:
When using additional custom CORS headers in Serverless, the local `serverless webpack serve` command does not allow those headers when developing locally. This fixes that.


jelder:
Is this a duplicate of #74?

HyperBrain:
@dolfelt  It would be great if you could comment in #135 if you use the serve functionality or not, so that we can decide how to proceed there.

HyperBrain:

PR #110, opened 2017-03-03
Fixes #107 Hook on "deploy function" command

### LINKED ISSUE #107: Single function deploy does not invoke the plugin
I do not know if this is a problem with Serverless' lifecycle events or if it is a problem within the webpack plugin. If I try to upload a single function during development, using `serverless deploy function -f auth` the webpack build is not triggered and Serverless tries to upload the whole project which exceeds the upload file size limit of 70MB.

When deploying the complete service with `serverless deploy` the plugin is invoked correctly and the resulting function code ZIP is about 17MB.

```
Serverless: Deploying function: auth...
Serverless: Packaging function: auth...
Serverless: Uploading function: auth (90.02 MB)...

  Serverless Error ---------------------------------------

     Request must be smaller than 69905067 bytes for the
     UpdateFunctionCode operation

     OS:                 win32
     Node Version:       4.3.1
     Serverless Version: 1.6.1

```

If it turns out that the problem is on Serverless' side, we should create an issue there.

HyperBrain:
Fixes #107: This fix hooks the webpack plugin onto the Serverless "deploy function" command. Currently the plugin will webpack the service as it would be done with a regular deploy. Then the single selected function is deployed by Serverless.
This makes the whole thing work. However there could be some improvements done later. Maybe _this.options.f_ can be used in this case to restrict the webpack run to create only the selected function entry point. But finding that would not be trivial.
I suggest to approve this solution on the first hand and handle the rest as improvement.

### DIFF
--- a/index.js
+++ b/index.js
@@ -99,6 +99,15 @@ class ServerlessWebpack {
       'after:deploy:createDeploymentArtifacts': () => BbPromise.bind(this)
         .then(this.cleanup),
 
+      // This hook is exposed by SLS' deploy function command
+      'before:deploy:function:packageFunction': () => BbPromise.bind(this)
+        .then(this.validate)
+        .then(this.compile)
+        .then(this.packExternalModules),
+
+      'after:deploy:function:packageFunction': () => BbPromise.bind(this)
+        .then(this.cleanup),
+
       'webpack:validate': () => BbPromise.bind(this)
         .then(this.validate),
 


mnightingale:
This doesn't seem to work when you have multiple entry/outputs in webpack.config.js similar to using the example config from https://webpack.github.io/docs/multiple-entry-points.html

> $ sls deploy function -f x
> Serverless: Deploying function: x...
> Serverless: Bundling with Webpack...
> ts-loader: Using typescript@2.2.1 and /Users/XYZ/tsconfig.json
> Time: 1248ms
>       Asset     Size  Chunks             Chunk Names
>     x.js   620 kB       0  [emitted]  x
>    y.js   4.5 kB       1  [emitted]  y
>  z.js  3.21 kB       2  [emitted]  z
> Serverless: Packaging function: x...
> path.js:7
>     throw new TypeError('Path must be a string. Received ' + inspect(path));
>     ^
> 
> TypeError: Path must be a string. Received undefined
>     at assertPath (path.js:7:11)
>     at Object.basename (path.js:1355:5)
>     at fse.copy (/Users/XYZ/node_modules/serverless-webpack/lib/cleanup.js:37:22)
>     at doneOne (/Users/XYZ/node_modules/fs-extra/lib/copy/ncp.js:237:40)
>     at /Users/XYZ/node_modules/fs-extra/lib/copy/ncp.js:122:11
>     at /Users/XYZ/node_modules/graceful-fs/polyfills.js:239:18
>     at FSReqWrap.oncomplete (fs.js:112:15)
>

martinmicunda:
@mnightingale The error is caused by `after:deploy:function:packageFunction` if you remove that part of the code it works however I found another problem when you run `sls deploy function -f test` it will create `.serverless` folder under `.webpack` directory so structure looks like this:

```
|-.webpack
    |-.serverless
         |-test.zip
```

and this cause follow error as the serverless expact zip file to be under `.serverless` in root folder not in `.webpack`:

**ENOENT: no such file or directory, open '<-your-path->/.serverless/test.zip'**

My current solution is below but is not ideal however it still better then deploy cloudformation during development:

```bash
sls package -v && sls deploy function -f test -v
```

franciscocpg:
Hello @martinmicunda .
This PR https://github.com/elastic-coders/serverless-webpack/pull/130 should fix it.

HyperBrain:
Superseded by #130

PR #112, opened 2017-03-20
sls webpack serve / Fix Response not returning Headers

aheissenberger:
AWS Lamda supports setting Headers of Response

Use this to test the change https://github.com/elastic-coders/serverless-webpack/issues/84#issuecomment-287801245


HyperBrain:
@aheissenberger It would be great if you could comment in #135 if you use the serve functionality or not, so that we can decide how to proceed there.

HyperBrain:

PR #129, opened 2017-06-02
Add support for webpack configuration in TypeScript format

jussikinnula:
To add support for webpack configuration in TypeScript format (e.g. webpack.config.ts), there's an existing package `ts-node` which just does the magic.

Webpack uses this method also internally, if specifying configuration file to webpack by using `webpack --config webpack.config.ts`. But since `serverless-webpack` does require the configuration  by using a  "require", `ts-node` needs to be required to support on-the-fly `.ts` -file requires.

### DIFF
--- a/index.js
+++ b/index.js
@@ -15,6 +15,15 @@ class ServerlessWebpack {
     this.serverless = serverless;
     this.options = options;
 
+    if (
+      this.serverless.service
+      && this.serverless.service.custom
+      && this.serverless.service.custom.webpack
+      && this.serverless.service.custom.webpack.endsWith('.ts')
+    ) {
+      require('ts-node/register');
+    }
+
     Object.assign(
       this,
       validate,
--- a/package.json
+++ b/package.json
@@ -31,7 +31,8 @@
     "body-parser": "^1.15.2",
     "express": "^4.14.0",
     "fs-extra": "^0.26.7",
-    "npm-programmatic": "0.0.5"
+    "npm-programmatic": "0.0.5",
+    "ts-node": "^3.2.0"
   },
   "devDependencies": {
     "chai": "^3.5.0",


hassankhan:
Shouldn't we detect whether a TypeScript format configuration is being used before loading `ts-node`?

HyperBrain:
I agree with @hassankhan . Unconditionally requiring `ts-node` and triggering all its side-effects maybe unwanted if you don't use TypeScript at all.
What about adding a command line switch `--ts` to conditionally enable TypeScript support on demand and initialize (require) it only then?

hassankhan:
I was thinking of something more lo-fi like checking the file extension, either way sounds good 👍

jussikinnula:
I originally had the `require('ts-node/register');` conditionally, just before starting webpack compilation. So I'll update the PR next week when I'm back at work :-) ...

I wouldn't worry about any side effects, since ts-node actually does hook into `require()` and `.ts` -files. But of course it's a consideration always to add any extra dependencies.

HyperBrain:
@jussikinnula Any update here?

jussikinnula:
Updated version of the patch is now on top of latest upstream master!

I tested with my Angular starter project, and works (e.g. `npm run sls:local`):
https://github.com/jussikinnula/angular-serverless-typescript

index.js
```diff
@@ -15,6 +15,15 @@ class ServerlessWebpack {
     this.serverless = serverless;
     this.options = options;
 
+    if (
+      this.serverless.service
```
hassankhan: Could these perhaps use Lodash's `has()` instead? Something like:

```js
if (
_.has(this.serverless.service, 'custom.webpack') 
&& this.serverless.service.custom.webpack.endsWith('.ts')
) {
...
}
```

HyperBrain: .. and lodash's `_.endswith()` 😄

hassankhan: Updated the comment 😉

hassankhan:
Aside from the one thing mentioned, this looks good to go! 👍

jussikinnula:
I didn't want to add any extra dependencies besides `ts-node`, and also I tried to follow the coding style used elsewhere. If you want I can add `lodash` and use it's patterns, but I guess it would be a separate thing since similar if -clauses exist in various places under `lib/`.

HyperBrain:
That's a valid point. We can do a "lodash"ification as a completely separate branch/PR and switch the code to use it everywhere, not only here. I will create a task for that, so that we get the whole codebase a bit leaner.
So this one is good to go.

PR #130, opened 2017-06-02
Integration with deploy package function

franciscocpg:
Fixes issue https://github.com/elastic-coders/serverless-webpack/issues/107.
Also PR https://github.com/elastic-coders/serverless-webpack/pull/110 seems incomplete and it looks like the author is not working on it anymore.
Here is a working `webpack.config.js`
```javascript
const yaml = require('js-yaml');
const fs = require('fs');
const path = require('path');
const nodeExternals = require('webpack-node-externals');

const handlerRegex = /\.[_$a-zA-Z\xA0-\uFFFF][_$a-zA-Z0-9\xA0-\uFFFF]*$/;
const include = './_webpack/include.js';
const entries = {};

const doc = yaml.safeLoad(fs.readFileSync('serverless.yml', 'utf8'));

// If we pass a function, package just it 
// not all the functions listed on serverless.yml
const serverlessFunction = process.env.serverlessFunction;
const functions = serverlessFunction ?
  { serverlessFunction: doc.functions[serverlessFunction] }
  : doc.functions;
for (const key in functions) { // eslint-disable-line guard-for-in
  const handler = functions[key].handler;
  const entryKey = handler.replace(handlerRegex, '');
  entries[entryKey] = [include, `./${entryKey}.js`];
}

module.exports = {
  entry: entries,
  target: 'node',
  // Generate sourcemaps for proper error messages
  devtool: 'source-map',
  // Since 'aws-sdk' is not compatible with webpack,
  // we exclude all node dependencies
  externals: [nodeExternals()],
  // Run babel on all .js files and skip those in node_modules
  module: {
    loaders: [{
      test: /\.js$/,
      loaders: ['babel-loader'],
      include: __dirname,
      exclude: /node_modules/,
    }]
  },
  output: {
    libraryTarget: 'commonjs',
    filename: '[name].js'
  }
};
```


afhammad:
Installed and tested this PR

Although Lambda shows this when opening the function on AWS console:

> Your Lambda function "fnName" cannot be edited inline since the file name specified in the handler does not match a file name in your deployment package.

The function does get packaged correctly and work.

franciscocpg:
@HyperBrain 
Related to https://github.com/elastic-coders/serverless-webpack/issues/134
This PR replaces your PR https://github.com/elastic-coders/serverless-webpack/pull/110.
I have used it in some of my projects for about 1 month now without any problems.

HyperBrain:
@franciscocpg Ok. Then I will close my PR.

HyperBrain:
Regarding the new label please check my comment https://github.com/elastic-coders/serverless-webpack/issues/134#issuecomment-312304294

HyperBrain:
@franciscocpg Can you check if #91 is also replaced, as it also about hooking package function and on the first sight does something quite similar?

franciscocpg:
@HyperBrain. Yes it replace. 
In fact I've tested the branch from https://github.com/elastic-coders/serverless-webpack/pull/91 and it's not working because the zip file is not being copied to `.serverless`.

lib/validate.js
```diff
@@ -18,6 +18,9 @@ module.exports = {
         throw new this.serverless.classes
           .Error('The webpack plugin could not find the configuration file at: ' + webpackConfigFilePath);
       }
+      if (this.options.function) {
+        process.env.serverlessFunction = this.options.function;
```
HyperBrain: Is there any reason why an additional environment variable is set here? I would not set anything in the process' environment as it is global. Better would be to set it as member of the plugin object with `this.XXX = ...` in case it is needed by the plugin itself.
If it is meant to be used in the vars of serverless.yml, you can reference it via options as `${opt:function}` there, even with specifying a default in case it is not set (`${opt:function, 'none'}`).

lib/copyFunctionArtifact.js
```diff
@@ -0,0 +1,31 @@
...
+
+      readStream.once('error', (err) => {
+        reject(error);
+      });
+
+      readStream.once('end', () => {
+        resolve();
+      });
+
+      readStream.pipe(fs.createWriteStream(artifactFilePath));
```
HyperBrain: Can this call itself throw, or does it delegate any errors to the 'error' event in any case?
Otherwise the pipe call itself should be wrapped into a try/catch and call reject() in the catch block.

johnf:
I like this approach. I wonder though if we should provide a utility function. Something like

``` yaml

module.exports = {
  entry: slsw.entries(),
  // other bits
  output: {
    libraryTarget: 'commonjs',
    filename: '[name].js'
  }
};
```

HyperBrain:
@johnf Interesting idea 🙌 . As this is out of scope for this particular PR, would you mind creating a separate issue for the "library extension" of the plugin with a detailed description what problem it addresses? Thanks.

hassankhan:
I think having exported entries is a pretty cool idea 👍 .

We don't really do single function deploys, so my experience is a bit limited 😅 . My only worry would be to inadvertently affect whole package deploys. That said, it would save a lot of trouble having to manually sync `serverless.yml` and `webpack.config.js` and I'd be up for pushing to `v2.1` so we can test it properly.

HyperBrain:
Ok. I'll tag it accordingly... we will do function deploys as soon as this is done 😄

lib/validate.js
franciscocpg: @HyperBrain I'm using this env var with `webpack.config.js`. But reading further comments I realized that you guys have suggested a better approach.

lib/copyFunctionArtifact.js
franciscocpg: @HyperBrain pipe emits an error event so the better approach would be:
```js
readStream.pipe(fs.createWriteStream(artifactFilePath)).on('error', (err) => {
  reject(error);
});
```

HyperBrain: Sounds reasonable 👍

HyperBrain:
@franciscocpg Do you need any help with the entries fix? I could create a separate branch in the repository to consolidate the feature, so that we can add this with a separate PR before it finds its way to master.

franciscocpg:
Hi @HyperBrain 
So if I don't misunderstood anything we are going to move this logic (with some modifications) from `webpack.config.js`
```javascript
const entries = {};

const doc = yaml.safeLoad(fs.readFileSync('serverless.yml', 'utf8'));

// If we pass a function, package just it 
// not all the functions listed on serverless.yml
const serverlessFunction = process.env.serverlessFunction;
const functions = serverlessFunction ?
  { serverlessFunction: doc.functions[serverlessFunction] }
  : doc.functions;
for (const key in functions) { // eslint-disable-line guard-for-in
  const handler = functions[key].handler;
  const entryKey = handler.replace(handlerRegex, '');
  entries[entryKey] = [include, `./${entryKey}.js`];
}
```
to our plugin (a new file called `index.js` or `lib.js`) and then export `entries` (through lib expose) so one could use like this `webpack.config.js`:
```js
module.exports = {
  entry: slsw.lib.entries,
  // other bits
  output: {
    libraryTarget: 'commonjs',
    filename: '[name].js'
  }
};
```
Did I understand correctly?

HyperBrain:
I understood it that way 😄 . The webpack config should have no knowledge of Serverless internals, so everything that interacts directly with the plugin or Serverless should be in the plugin.

After a solution is finsihed we also should test the behavior with all possible configurations (may it be the entries retrieved from the plugin or statically set entries).

franciscocpg:
Sorry for the late, guys. I was very busy these days.
I made the changes we've discussed here and also I've updated examples folder.
IMO `babel-dynamically-entries` (entries retrieved from the plugin) and `babel-multiple-statically-entries` (statically set entries) examples are a good place to do some manual tests for this PR (`sls deploy`, `sls deploy function -f first`, `sls deploy function -f second`, `sls webpack invoke -f first`, `sls webpack invoke -f second`).
I think this PR is still a WIP because I didn't add any specs to tests (hope to do this ASAP) but it looks like everything is working as expected.

HyperBrain:
@franciscocpg Great stuff and a big thank you 🙌 . I will test it over the weekend and review the PR again.

HyperBrain:
@franciscocpg Just testing now and it looks good so far. Could you rebase the PR onto the current master before you add the unit tests? I upgraded the test tools to the most recent versions.

HyperBrain:
Can you try with `git rebase` to rebase the branch onto the current master? I think the "merge" commit (the latest one) somehow messes up GitHub's diff display because it does not change the commit dates.

franciscocpg:
@HyperBrain 
Do you mean `git rebase master`, right?

HyperBrain:
Yes, given that your fork's master is up-to-date with the webpack's master 👍 .
But first remove the merge commit with "reset"

HyperBrain:
According to the git docs (https://git-scm.com/docs/git-rebase) it should be `git rebase --onto master`

HyperBrain:
I managed to rebase the commits locally. If you agree, I can commit them back to your branch into the PR.

franciscocpg:
Ok @HyperBrain

HyperBrain:
Can you check if "enable edits by maintainers" is checked in the PR?

franciscocpg:
I've just checked `Allow edits from maintainers` now

HyperBrain:
It worked 💯 

You can now pull the branch again locally. It is now rebased onto the current master.

examples/babel-dynamically-entries/package.json
```diff
@@ -0,0 +1,19 @@
...
+  "author": "Nicola Peduzzi <thenikso@gmail.com> (http://nikso.net)",
+  "license": "MIT",
+  "devDependencies": {
+    "babel-core": "^6.24.1",
+    "babel-loader": "^7.1.1",
+    "babel-plugin-transform-runtime": "^6.23.0",
+    "babel-polyfill": "6.13.0",
+    "babel-preset-env": "^1.6.0",
+    "serverless-webpack": "file:../../",
+    "webpack": "^1.13.1"
```
HyperBrain: As we support any webpack version now, maybe using webpack 2 in the examples would be good. What do you think?

franciscocpg: :+1: 
I'm going to update it

franciscocpg: I've updated all examples dependencies and tested all folders.
It looks like everything is working as expected.

franciscocpg:
@HyperBrain 
I've added tests (file `validate.test.js`).

HyperBrain:
I found an issue with handlers in arbitrary paths when testing with one of our bigger production projects.
I fixed it and added the commits. All my tests now worked properly.
Additionally I added lodash and used the serverless framework to retrieve the service functions which is the proposed way to do that.
This should be good to go now :)

franciscocpg:
Nice @HyperBrain! 
Just a note about this commit
* https://github.com/elastic-coders/serverless-webpack/commit/b4a9e1c Removed yarn lock file. yarn should use package-lock.json

It looks like **yarn** isn't using `package-lock.json` right now: https://github.com/yarnpkg/yarn/issues/3614
So if I ran `yarn` at the root of this repository, it's going to generate a `yarn.lock`.
According to the issue above right now we should choose to support `package-lock.json` (npm) **OR** `yarn.lock` (yarn). 
Maybe in the future we could use `yarn` with `package-lock.json`.
This thread may be helpful: https://stackoverflow.com/a/44904494

HyperBrain:
@franciscocpg Yeah, at the moment it seems to be quite a mess with the different package lock mechanisms. Maybe we should remove the lock files completely for now, and use it as soon as the tools (npm and yarn) can work with the same one.

franciscocpg:
@HyperBrain 
IMO we should remove `package-lock.json` and add `yarn.lock`.  Of course this means that we are "officially" migrating our node package manager to `yarn` (maybe we should open another issue for this discussion?).
But I don't think is a good ideia to remove lock files for now. I think we should decide if we migrate to `yarn` (adding `yarn.lock` and removing `package-lock.json`) or keep with `npm` as our official node package manager (keeping `package-lock.json`).

franciscocpg:
For this PR maybe we should keep `npm` and create a new issue about migration to `yarn` so community could contribute with opinions.
What do you think @HyperBrain ?

HyperBrain:
That sounds good to me. This PR should not invent or change things other than the problem solved ;-)
I will open a separate issue for the yarn/npm discussion.

PR #131, opened 2017-06-09
Add lifecycle hook for serverless offline start

Gregoirevda:
Serverless offline has 2 root commands to start : 
- `serverless offline` which triggers `offline:start` life cycle hook (This is working)
- `serverless offline start` which triggers `offline:start:init` and  `offline:start:end` life cycle hooks (This is NOT working)

The second one is important, because it has an event on start up and on end.
If you want to work with serverless-offline, serverless-webpack and serverless-dynamodb-local. You need to call `sls offline start` which will also stop the db and stop serverless offline

### DIFF
--- a/index.js
+++ b/index.js
@@ -123,6 +123,11 @@ class ServerlessWebpack {
       'before:offline:start': () => BbPromise.bind(this)
         .then(this.validate)
         .then(this.wpwatch),
+      
+      'before:offline:start:init': () => BbPromise.bind(this)
+        .then(this.validate)
+        .then(this.wpwatch),
+      
     };
   }
 }


Gregoirevda:
Event on serverless-dynamodb-local: https://github.com/99xt/serverless-dynamodb-local/blob/v1/index.js#L94
Event on serverless-offline: https://github.com/dherault/serverless-offline/blob/master/src/index.js#L117

jerep6:
+1

Duplicate but more complete than the PR #105

HyperBrain:
I will test this locally with one of our projects (that is invoked with `offline` without `start`) to make sure that it does not break this use case. As soon as that works I'll approve it 😄

HyperBrain:
Tested -> works without specifying `start` as before. LGTM from my side.
/cc @hassankhan

PR #153, opened 2017-07-14
Integrate with "serverless invoke local"

### LINKED ISSUE #151: Integrate with "Serverless invoke local" and change the "webpack invoke" command
# This is a Feature Proposal

## Description

Serverless offers a fully working `local invoke` command that handles all possible
kinds of invocations, sets the function's environment correctly and much more.

There is no need to have a separate `webpack invoke` command that replicates (only)
parts of this functionality and binds resources to maintain it properly.

It would be better to integrate seamlessly with `local invoke` by hooking the important
lifecycle events exposed by the command. Finally, `serverless invoke local` should 
invoke the function after the webpack compile has been triggered as part of the lifecycle
hooks in place.

To make it completely seamless, the `webpack invoke` command should techincally just use Serverless'
pluginManager to spawn the `invoke local` command like this: `this.serverless.pluginManager.spawn('invoke:local')`.

This solution would also make the invocation provider independent as Serverless fully takes care of that already.

Related issue(s):
#128 (arbitrary provider support)

HyperBrain:
## What did you implement:

Closes #151 

### SLS invoke local / webpack invoke
Hook into `serverless invoke local` and emit an error if the old deprecated `webpack invoke` is used.

This has a lot of advantages over the old self-implemented command. Some of them are:
Other plugins will be used too, the environment is set properly, input events can be specified and
all of the functions offered by Serverless are available.

### webpack watch
Changed the watch implementation to use "serverless invoke local" to run the code. Now all switches (-p|--path, -d|--data) are possible.

## How can we verify it:

Use `serverless invoke local` or `serverless invoke local --watch` the same was as you do with any project that does not use `serverless-webpack`

## Todos:

- [x] Write tests
- [x] Write documentation


HyperBrain:
Webpack watch now also uses "invoke local" under the hood. Maybe it would be better if I substitute `webpack watch` with `serverless invoke local --watch ...`. I think having a separate new switch for invoke local is much more convenient as users do not need to remember the special webpack commands anymore.
@hassankhan @benjaminwood What do you think?

benjaminwood:
@HyperBrain I think that's an excellent idea. 👍

HyperBrain:
`webpack watch` is removed now in favor of `serverless invoke local ... --watch` as I said in my previous comment.

benjaminwood:
Looking good! 🙌

PR #155, opened 2017-07-14
fix serverless-offline exec

### LINKED ISSUE #154: serverless-offline exec is broken
# This is a Bug Report

## Description

Serverless-offline has a nifty feature called scoped execution that can be used for automated integration testing.
* https://www.npmjs.com/package/serverless-offline#scoped-execution

This exec feature is currently not working with serverless-webpack because the tests get executed before webpack watch has a chance to compile the source.

## Proposed Solution

The proposed solution is to execute compile after validate and before watch. This will force a compile when serverless-offline is started but before the tests are executed and still support watch for non automated scenarios that do not have the race condition.

## Additional Data

* ***Serverless-Webpack Version 2.0.0***:
* ***Webpack version 3.0.0***:
* ***Serverless Framework Version 1.17.0***:
* ***Serverless-Offline Version 3.15.0***:

jgilbert01:
## What did you implement:

Closes #154

## How did you implement it:

Call compile on serverless-offline start after validate but before watch.

## How can we verify it:

`sls offline start --exec "./startIntegrationTests.sh"`

* https://github.com/dherault/serverless-offline#scoped-execution

### DIFF
--- a/index.js
+++ b/index.js
@@ -122,10 +122,12 @@ class ServerlessWebpack {
 
       'before:offline:start': () => BbPromise.bind(this)
         .then(this.validate)
+        .then(this.compile)
         .then(this.wpwatch),
       
       'before:offline:start:init': () => BbPromise.bind(this)
         .then(this.validate)
+        .then(this.compile)
         .then(this.wpwatch),
       
     };
--- a/lib/compile.js
+++ b/lib/compile.js
@@ -23,12 +23,10 @@ module.exports = {
         if (stats.compilation.errors.length) {
           throw new Error('Webpack compilation error, see above');
         }
-        if (!this.options.location) {
-          const outputPath = stats.compilation.compiler.outputPath;
-          this.webpackOutputPath = outputPath;
-          this.originalServicePath = this.serverless.config.servicePath;
-          this.serverless.config.servicePath = outputPath;
-        }
+        const outputPath = stats.compilation.compiler.outputPath;
+        this.webpackOutputPath = outputPath;
+        this.originalServicePath = this.serverless.config.servicePath;
+        this.serverless.config.servicePath = outputPath;
         return stats;
       });
   },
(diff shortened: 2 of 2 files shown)


lib/compile.js
```diff
@@ -23,10 +23,12 @@ module.exports = {
...
-        const outputPath = stats.compilation.compiler.outputPath;
-        this.webpackOutputPath = outputPath;
-        this.originalServicePath = this.serverless.config.servicePath;
-        this.serverless.config.servicePath = outputPath;
+        if (!this.options.location) {
+          const outputPath = stats.compilation.compiler.outputPath;
+          this.webpackOutputPath = outputPath;
+          this.originalServicePath = this.serverless.config.servicePath;
+          this.serverless.config.servicePath = outputPath;
+        }
```
jgilbert01: * With adding the compile call the serverless-offline location option is no longer necessary
* For backwards compatibility we can skip these lines when the location option is present

index.js
```diff
@@ -122,10 +122,12 @@ class ServerlessWebpack {
 
       'before:offline:start': () => BbPromise.bind(this)
         .then(this.validate)
+        .then(this.compile)
         .then(this.wpwatch),
       
       'before:offline:start:init': () => BbPromise.bind(this)
         .then(this.validate)
+        .then(this.compile)
         .then(this.wpwatch),
```
jgilbert01: * this forces a compile at start up that completes before offline exec us called
* watch remains to support manual scenarios

lib/compile.js
HyperBrain: I'm not sure if changing the serverless servicePath here will have side-effects on all other functions. The problem is, that Serverless uses this path for it's package command, persists it's service state there, which it reloads when entering the deploy phase. This change does not only affect the offline use case, but also the package/deploy use cases.

Other plugins rely on the service path to be set as Serverless sets it.

I's propose to remove this change from the PR (it is imo not necessary to fix the race condition).

jgilbert01: sounds good to me

Issue #108, opened 2017-02-27
Getting sourcemaps to work


ajmath:
What is the trick for getting sourcemaps to work correctly with the stacktraces in AWS logs?

I've currently got `require('source-map-support').install();` after all other imports in a file `/src/lambdas/webhook.ts`.

Webpack config looks like this:
```
var path = require('path');

module.exports = {
  entry: {
    '/src/lambdas/webhook': './src/lambdas/webhook.ts',
    '/src/lambdas/queue-poller': './src/lambdas/queue-poller.ts',
    '/src/lambdas/queue-worker': './src/lambdas/queue-worker.ts',
  },
  target: 'node',
  devtool: 'source-map',
  resolve: {
    modules: [
      path.join(__dirname, 'src'),
      path.join(__dirname, 'node_modules')
    ],
    extensions: ['', '.ts', '.js']
  },
  module: {
    loaders: [
      { test: /\.ts(x?)$/, loader: 'ts-loader' }
    ]
  },
  resolve: {
    extensions: ['.ts', '.js'],
  },
  output: {
    libraryTarget: 'commonjs',
    path: path.join(__dirname, '.webpack'),
    filename: '[name].js'
  },
};
```

Here is the contents of the uploaded zip:
```
 ⇒ unzip -l .serverless/trebuchet.zip
Archive:  .serverless/trebuchet.zip
  Length      Date    Time    Name
---------  ---------- -----   ----
     3020  02-24-2017 22:37   src/lambdas/queue-poller.js
     3140  02-24-2017 22:37   src/lambdas/queue-poller.js.map
     2956  02-24-2017 22:37   src/lambdas/queue-worker.js
     3073  02-24-2017 22:37   src/lambdas/queue-worker.js.map
  3655773  02-24-2017 22:37   src/lambdas/webhook.js
  5682470  02-24-2017 22:37   src/lambdas/webhook.js.map
---------                     -------
  9350432                     6 files
```
The output for both logged exceptions and unhandled exceptions is still using the full bundle.

Am I missing something from the `source-map-support` config?  

If everything is working correctly should the logged and unhandled exceptions be using the stacktraces from the sourcemap?

benjaminwood:
Has anybody figured out how to get [source-map-support](https://github.com/evanw/node-source-map-support) to work? In my case I'd like source-maps to be used in the trace when I invoke a function locally. As shown in the readme: `serverless webpack invoke --function <function-name>`

[...]

PR #156, opened 2017-07-15
Make processed input path absolute

benjaminwood:
## What did you implement:

Related: #151

When using the invoke local command with the --path (-p) flag, relative file paths will not be found because serverless will end up looking in the build directory later on. So, this PR ensures that the path is absolute, ensuring it will be found, even if the path was relative.

## How did you implement it:

I added a function in utils.js that gets the --path option if present. It then uses `path.resolve` to resolve it, then assigns it back to serverless config.

## How can we verify it:

Just use `invoke local` with the --path option:

`serverless invoke local -f function_name --path data.json`

## Todos:

- [ ] Write tests
- [ ] Write documentation

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -1,8 +1,5 @@
 'use strict';
 
-const BbPromise = require('bluebird');
-const path = require('path');
-
 function guid() {
   function s4() {
     return Math.floor((1 + Math.random()) * 0x10000)
@@ -35,18 +32,8 @@ function searchCache(moduleName, callback) {
   }
 }
 
-function makeProcessedInputPathAbsolutePath() {
-  var originalPath = this.serverless.config.serverless.processedInput.options.path
-  if (originalPath) {
-    var absolutePath = path.resolve(originalPath)
-    this.serverless.config.serverless.processedInput.options.path = absolutePath
-  }
-  return BbPromise.resolve();
-}
-
 module.exports = {
   guid,
   purgeCache,
-  searchCache,
-  makeProcessedInputPathAbsolutePath
+  searchCache
 };
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -1,5 +1,8 @@
 'use strict';
 
+const BbPromise = require('bluebird');
+const path = require('path');
+
 function guid() {
   function s4() {
     return Math.floor((1 + Math.random()) * 0x10000)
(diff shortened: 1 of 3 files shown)


benjaminwood:
I'm thinking that `makeProcessedInputPathAbsolutePath` may not belong in utils. Should I create a file for this in the spirit of compile and validate? It'd be more consistent that way, it just seems like such a small thing, limited usefulness to `invoke local`... thoughts?

HyperBrain:
Yes, agree. `makeProcessedInputPathAbsolutePath` is not an utility function. It is a execution building block needed to execute `invoke local`. So an own file would be the best solution.

lib/utils.js
```diff
@@ -32,8 +35,18 @@ function searchCache(moduleName, callback) {
   }
 }
 
+function makeProcessedInputPathAbsolutePath() {
+  var originalPath = this.serverless.config.serverless.processedInput.options.path
```
HyperBrain: Should be `const` instead of `var`. Also a missing semicolon.

lib/utils.js
```diff
@@ -32,8 +35,18 @@ function searchCache(moduleName, callback) {
...
+  if (originalPath) {
+    var absolutePath = path.resolve(originalPath)
```
HyperBrain: `const` and semicolon.

Did you check this for a given absolute path (i.e. if originalPath is already absolute) too?

benjaminwood:
@HyperBrain thanks, I've made some revisions:

1. Rename the function to: `makePathOptionAbsolute`.
2. Move the function to its own file in lib `makePathOptionAbsolute.js`.
3. Made use of `const` and added missing semicolons.

Let me know if there is anything else I can do!

lib/utils.js
benjaminwood: > Did you check this for a given absolute path (i.e. if originalPath is already absolute) too?

Yes, this works with absolute URLS as well.

```
> path.resolve('foo')
'/Users/Ben/js_projects/serverless-webpack/foo'
> path.resolve('./foo')
'/Users/Ben/js_projects/serverless-webpack/foo'
> path.resolve('/Users/Ben/js_projects/serverless-webpack/foo')
'/Users/Ben/js_projects/serverless-webpack/foo'
```

HyperBrain:
@benjaminwood It's good to go now 🙌 . I'll merge it into the feature branch and continue with the finalization. Then we have a PR for the complete feature onto master and can do a final test and review.

I'll let you know if there is anything you can jump in - just thought about something 😄 : If you don't mind, you could dive a bit into the source-map stuff mentioned in #108 and check how sourcemaps can be used properly with local invoke. If they work, we should create a PR to update the documentation (README) and show users how to setup a working system.

benjaminwood:
Thanks @HyperBrain! I'll take a look at #108.

PR #167, opened 2017-07-27
Do not add .js extension to entries values

### LINKED ISSUE #165: Can't use the new slsw.lib.entries with typescript
# This is a Bug Report

## Description

`slsw.lib.entries` returns an object with keys `name.js`, and as values the string `./name.js`.
This is making developing with typescript quite difficult:
1. The key shouldn't have an extension in case i'm using an output name like `[name].[hash].js`, or in case I actually want a different extension.
2. The value should't have an extension as well, so I can actually input typescript files

## Additional Data

* ***Serverless-Webpack Version you're using***: 2.1.0

kikar:
## What did you implement:

Closes #165 

## Todos:

- Write tests
- [x] Write documentation
- Provide verification config / commands / resources
- Enable "Allow edits from maintainers" for this PR
- Update the messages below

### DIFF
--- a/lib/validate.js
+++ b/lib/validate.js
@@ -3,16 +3,38 @@
 const BbPromise = require('bluebird');
 const path = require('path');
 const fse = require('fs-extra');
+const fs = require('fs');
 const lib = require('./index');
 const _ = require('lodash');
 
-function getEntryForFunction(serverlessFunction) {
+function listFiles(dir, filelist) {
+  const files = fs.readdirSync(dir);
+  filelist = filelist || [];
+  files.forEach(function(file) {
+    if (fs.statSync(path.join(dir, file)).isDirectory()) {
+      filelist = listFiles(path.join(dir, file), filelist);
+    }
+    else {
+      filelist.push(file);
+    }
+  });
+  return filelist;
+}
+
+function getEntryExtension(fileName, files) {
+  const fileFullName = files.find(file => file.startsWith(fileName));
+  return path.extname(fileFullName);
+}
+
+function getEntryForFunction(serverlessFunction, files) {
 	const handler = serverlessFunction.handler;
-	const handlerFile = /(.*)\..*?$/.exec(handler)[1];
+  const handlerFile = /(.*)\..*?$/.exec(handler)[1];
+  
+  const ext = getEntryExtension(handlerFile, files);
 
 	// Create a valid entry key
 	return {
-		[handlerFile]: `./${handlerFile}`
+		[handlerFile]: `./${handlerFile}${ext}`
 	};
 };
 
@@ -27,13 +49,14 @@ module.exports = {
 		// Expose entries - must be done before requiring the webpack configuration
     const entries = {};
     const functions = this.serverless.service.getAllFunctions();
+    const allFilesInServicePath = listFiles(this.serverless.config.servicePath);
     if (this.options.function) {
 			const serverlessFunction = this.serverless.service.getFunction(this.options.function);
-      const entry = getEntryForFunction(serverlessFunction);
+      const entry = getEntryForFunction(serverlessFunction, allFilesInServicePath);
       _.merge(entries, entry);
     } else {
       _.forEach(functions, func => {
-        const entry = getEntryForFunction(this.serverless.service.getFunction(func));
+        const entry = getEntryForFunction(this.serverless.service.getFunction(func), allFilesInServicePath);
         _.merge(entries, entry);
       });
     }
--- a/lib/validate.js
+++ b/lib/validate.js
@@ -7,6 +7,17 @@ const glob = require('glob');
 const lib = require('./index');
 const _ = require('lodash');
 
+/**
+ * For automatic entry detection we sort the found files to solve ambiguities.
+ * This should cover most of the cases. For complex setups the user should
+ * build his own entries with help of the other exports.
+ */
+const preferredExtensions = [
+  '.js',
+  '.ts',
+  '.jsx'
+];
+
 module.exports = {
   validate() {
     const getEntryExtension = fileName => {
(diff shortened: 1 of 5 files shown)


lib/validate.js
```diff
@@ -8,7 +8,7 @@ const _ = require('lodash');
 
 function getEntryForFunction(serverlessFunction) {
 	const handler = serverlessFunction.handler;
-	const handlerFile = /(.*)\..*?$/.exec(handler)[1] + '.js';
+	const handlerFile = /(.*)\..*?$/.exec(handler)[1];
```
HyperBrain: We should actually retrieve the correct file extension here and set it in the value of the key, so that there's no need to use custom code in the webpack configuration to get the plugin running.

```
function getHandlerFileExtension(baseName) {
  // Use fs/fs-extra/path or any other appropriate mechanism to determine the actual file extension of the handler
  // Maybe we can search for known and supported endings in a predefined order
  // So that the standard extension 'js' is preferred if, by accident, there are multiple copies
  // of the handler with different extensions - or throw an error in that case.
  const allowedExtensions = [
    'js', 'ts', 'jsx'
  ];
  const handlerExtension = ...;
  ...
  return handlerExtension;
}
...
const handlerFileBaseName = /(.*)\..*?$/.exec(handler)[1];
const handerFile = `${handlerFileBaseName}.${getHandlerFileExtension(handlerFileBaseName)}`;
...
return {
  [handlerFileBaseName]: `./${handlerFile}`
};
```

tests/validate.test.js
```diff
@@ -277,10 +277,10 @@ describe('validate', () => {
         .then(() => {
           const lib = require('../lib/index');
           const expectedLibEntries = {
-            'module1.js': './module1.js',
-            'module2.js': './module2.js',
-            'handlers/func3/module2.js': './handlers/func3/module2.js',
-            'handlers/module2/func3/module2.js': './handlers/module2/func3/module2.js',
+            'module1': './module1',
```
HyperBrain: The tests have to be adapted AFTER the changes above, because the used external functions to determine the extension have to be stubbed with sinon.

HyperBrain:
Dtermining the extension will also support the case where single functions in a project are updated and the used extension changes. E.g. if you change a function from js to jsx or from js to ts. Then the next deploy will just work out of the box because your config file does not need to know anything about the specific function configurations.

lib/validate.js
kikar: Do we assume all the function files are in the servicePath folder?

HyperBrain: At least somewhere under the servicePath. The handlerBaseName contains the path elements relative to the handler. So, for example, it could be `handlers/shop/order/handler`, but we can assume that handler files are never outside the service folder (of course the user can use symlinks to access files outside the service directory, but checking if the baseName + extension exists, will automatically reveil the correct results in this case)

HyperBrain: Better rename the `handlerFileBaseName` variable to `handlerFilePathWithBaseName` as it actually contains the path + the basename.

kikar:
Latest commit should do the job. Problem is that all tests fail cause we don't actually have files to check

lib/validate.js
```diff
@@ -27,13 +49,14 @@ module.exports = {
 		// Expose entries - must be done before requiring the webpack configuration
     const entries = {};
     const functions = this.serverless.service.getAllFunctions();
+    const allFilesInServicePath = listFiles(this.serverless.config.servicePath);
```
HyperBrain: Is it necessary to browse the whole service dir file tree here? I think that is very expensive - the directory might include node_modules etc.

Assuming that the handlers we have already point to the handler files we check, we know exactly where the files have to be. So just searching in getEntryExtension(handlerFile) for `handlerFile.*` should retrieve the full file automatically (and even shows us if it is ambiguous, which is an error on the user side if multiple of our allowed extensions are found). We do not need to be aware of the complete file hierarchy here.

HyperBrain:
I can take care of the unit tests as soon as we have a performant and working solution.

kikar:
Improved the file search as recommended

lib/validate.js
```diff
@@ -3,16 +3,24 @@
 const BbPromise = require('bluebird');
 const path = require('path');
 const fse = require('fs-extra');
+const glob = require('glob');
 const lib = require('./index');
 const _ = require('lodash');
 
-function getEntryForFunction(serverlessFunction) {
+function getEntryExtension(fileName, servicePath) {
+  let files = glob.sync(`${fileName}.*`, { cwd: servicePath });
```
HyperBrain: can be `const`

lib/validate.js
```diff
@@ -3,16 +3,24 @@
...
+  return path.extname(files[0]);
```
HyperBrain: We could emit a warning in case more than one file is found here:
```
if (_.size(files) > 1) {
  this.serverless.cli.log(`WARNING: More than one matching handlers found for ${fileName}. Using ${_.first(files)}.`);
}
```

Of course the now global function(s) have to be moved into validate() to have access to the correct `this`.

Also use `_.first(files)` here instead of `files[0]` as the lodash variant will not crash if no file is found.

HyperBrain:
I will add the unit tests and the doc/example updates tomorrow, so that we can release 2.2.0 also tomorrow.

HyperBrain:
I tested it with one of our projects. That worked nicely. Also rebased onto the current webpack master.
Should be everything ok now. The only thing that can happen is, that if there are unrelated files besides the handler, that just have a different ending, the glob could match them (e.g. `handler.doc`). However it will show the warning we implemented.

If that turns out to be a problem later, we can restrict the allowed extensions by setting the glob pattern to `${fileName}.@(js|ts|jsx)`.

HyperBrain:
Just added some more complex test cases and a preference sorting. I will do some final tests now.
For most standard uses of lib.entries it should work now - if someone needs more sophisticated handling he can program the complete stuff in his webpack.conf - thanks to the serverless and option exports.

PR #168, opened 2017-07-27
Expose service and options for dynamicity

### LINKED ISSUE #158: Is it possible to dynamically configure the WebPack configuration (e.g. to do Tree Shaking)?
# This is a Question

## Description
How can I tell webpack to do some Tree Shaking? When I import modules, I might not want the whole thing, but just a function, usually I would just execute webpack in production mode with the -p option, but doesn't seem there is a way to do this here.

kikar:
## What did you implement:

Closes #158 

## Todos:

- [x] Write tests
- [x] Write documentation
- Provide verification config / commands / resources
- Enable "Allow edits from maintainers" for this PR
- Update the messages below

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,7 +51,14 @@ module.exports = {
 };
 ```
 
-You can also let the plugin determine the correct handler entry points at build time.
+### serverless-webpack lib export helper
+
+serverless-webpack exposes a lib object, that can be used in your webpack.config.js
+to make the configuration easier and to build fully dynamic configurations.
+
+#### Automatic entry resolution
+
+You can let the plugin determine the correct handler entry points at build time.
 Then you do not have to care anymore when you add or remove functions from your service:
 
 ```js
@@ -81,6 +88,25 @@ module.exports = {
 };
 ```
 
+#### Full customization (for experts)
+
+The lib export also provides the `serverless` and `options` properties, through
+which you can access the Serverless instance and the options given on the command-line.
+
+This enables you to have a fully customized dynamic configuration, that can evaluate
+anything available in the Serverless framework. There are really no limits.
+
+Samples are: The current stage and the complete service definition. You thereby
+have access to anything that a Serverless plugin would have access to.
+
+Both properties should be handled with care and should never be written to,
+as that will modify the running framework and leads to unpredictable behavior!
+
+If you have cool use cases with the full customization, we might add your solution
+to the plugin examples as showcase.
+
+### Output
+
 Note that, if the `output` configuration is not set, it will automatically be
 generated to write bundles in the `.webpack` directory. If you set your own `output`
 configuration make sure to add a [`libraryTarget`][link-webpack-libtarget]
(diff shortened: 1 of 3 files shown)


kikar:
@HyperBrain opened a PR, could you help me with the documentation and testing?

HyperBrain:
Will have a look later today and review the changes 😃

kikar:
I realised now that if we expose the serverless and the option, we dont really need the entries anymore.
Does it make sense to remove it?

HyperBrain:
Not really. Entries contain the entry points needed for the actual Serverless run. It will also honor single function deployments. We should not move this functionality (that guarantees that it is set correctly) to the webpack configuration. Later on (when we fully support individual packaging, so that tree shaking is applied via multi-compile to let each function be optimized separately) entries must be completely under the control of the plugin.

With the new service and options export, it is in the developer's hand, if he wants to change the default behavior - with the risk of breaking things or over-engineering :) But the default behavior should be as simple as possible, so that any user can just enable the plugin, use lib.entries and is done.

kikar:
Example of my `webpack.config.js`.
Thanks to the `servicePath` it can stay in the root folder and not in every service.
```javascript
const { join } = require('path');
const UglifyPlugin = require('webpack').optimize.UglifyJsPlugin;

let entry = slsw.lib.entries;
for (let key in entry) {
  entry[key] = entry[key] + '.ts';
}

let servicePath = slsw.lib.serverless.config.servicePath;
let plugins = [];

if (slsw.lib.options.stage === 'prod') {
  plugins.push(new UglifyPlugin()); // Tree Shaking FTW!
}

module.exports = {
  entry,
  target: 'node',
  module: {
    loaders: [
      { test: /\.ts(x?)$/, loader: 'ts-loader' }
    ]
  },
  resolve: {
    extensions: ['.ts', '.js']
  },
  output: {
    path: join(slsw.lib.serverless.config.servicePath, '.webpack'),
    libraryTarget: 'commonjs',
    filename: '[name].js'
  },
  plugins
}
```

HyperBrain:
This looks good.

> let entry = slsw.lib.entries;
> for (let key in entry) {
>   entry[key] = entry[key] + '.ts';
> }

This will be obsolete if we do a file search and determine the correct handler extension in #167 instead of forcing the user to add code to his webpack conf to get it run.

In general, I think that your example (after the extension changes in #167 have been done) can be added to the examples folder as (typescript-with-dynamic configuration).

I will add the missing unit tests here and merge this after the other PR.

kikar:
> This will be obsolete if we do a file search and determine the correct handler extension in #167 instead of forcing the user to add code to his webpack conf to get it run.
I think that it can get complicated, but also gives less space for customisation. Imagine a situation where the developer wants to write a function in .ts and one in .js

It's only 4 lines of code that we're talking about in the end.

HyperBrain:
But everyone has to add them - exactly in the same way. The framework should do that properly (and can do that because it is deterministic).

HyperBrain:
I added unit tests and will update the README too. Then this should be good to go 😃 .

PR #177, opened 2017-07-30
Support individual packaging with multi-compile

### LINKED ISSUE #116: Some modules are missing when using webpack 2
# Problem
We tried to use Webpack 2 and the current master branch of the webpack plugin.
The bundle generation is creating a correct bundle, but the packaged node modules,
that are installed as sibling to the bundle by the plugin are missing some external modules.

I.e. although they are present in the local services' package.json, they do not appear in
the package.json that is temporarily created by the webpack plugin to install the needed
external modules.

This leads to "required module not found" errors, when executing the compiled code on AWS
after Serverless deployed it.

# Root cause
The plugin parses the webpack stats to find external modules by matching the returned stats
with "external ".

See comment below: Webpack 2's dependency optimization sometimes removes first level dependencies but keeps the removed dep's second level dependencies.

### LINKED ISSUE #117: Bundling of node modules works oddly
I use `webpack-node-externals` for bundling mode modules like this:
```
externals: [nodeExternals({
    modulesFromFile: true,
  })],
```
(whole config is here: https://github.com/keboola/developer-portal/blob/master/webpack.config.js)

And it worked quite well until recently when I decided to move the deployment to Docker and noticed that the code stopped working because some modules are missing. I have found out that the `nodeExternals()` method returns list of 47 modules however when I run `serverless deploy`, it bundles 137 modules. Although quite more then I want but the code still works. But the weird thing is that when I run the same deployment from a docker container (using this Dockerfile: https://github.com/keboola/developer-portal/blob/docker-travis/Dockerfile) it bundles only 18 modules, so the code fails on missing modules. Would you have any idea why is this going on? Thanks.

### LINKED ISSUE #120: serverless-webpack is not compatible with new function level individual packaging
I am using serverless 1.11.0 with the new function level individual packaging feature.
Here is a piece of my serverless.yml.

```
request:
   handler: request.handler
   package:
       individually: true
       include:
          - request.js
```

When I `sls deploy`, I get this error.

```
Serverless: Packaging service...

     ENOENT: no such file or directory, open '/Users/deathweasel/Documents/repos/****/****/.webpack/.serverless/service-functionname.zip'****

     For debugging logs, run again after setting the "SLS_DEBUG=*" environment variable.

  Stack Trace --------------------------------------------
Error: ENOENT: no such file or directory, open '/Users/deathweasel/Documents/repos/****/****/.webpack/.serverless/service-functionname.zip'****
    at Error (native)
    at Object.fs.openSync (fs.js:640:18)
    at Object.fs.readFileSync (fs.js:508:33)
    at AwsCompileFunctions.compileFunction (/Users/deathweasel/Documents/repos/****/****/node_modules/serverless/lib/plugins/aws/deploy/compile/functions/index.js:176:24)
    at serverless.service.getAllFunctions.forEach (/Users/deathweasel/Documents/repos/****/****/node_modules/serverless/lib/plugins/aws/deploy/compile/functions/index.js:218:39)
    at Array.forEach (native)
    at AwsCompileFunctions.compileFunctions (/Users/deathweasel/Documents/repos/****/****/node_modules/serverless/lib/plugins/aws/deploy/compile/functions/index.js:218:8)
    at BbPromise.reduce (/Users/deathweasel/Documents/repos/****/****/node_modules/serverless/lib/classes/PluginManager.js:210:55)
From previous event:
    at PluginManager.invoke (/Users/deathweasel/Documents/repos/****/****/node_modules/serverless/lib/classes/PluginManager.js:210:22)
    at PluginManager.run (/Users/deathweasel/Documents/repos/****/****/node_modules/serverless/lib/classes/PluginManager.js:225:17)
    at Serverless.run (/Users/deathweasel/Documents/repos/****/****/node_modules/serverless/lib/Serverless.js:97:31)
    at serverless.init.then (/Users/deathweasel/Documents/repos/****/****/node_modules/serverless/bin/serverless:23:50)
```

I did a quick examination of the code and it looks like serverless-webpack doesn't examine that setting at all.

### LINKED ISSUE #173: Use the standard Serverless servicePath in the plugin
# This is a Feature Proposal

## Description

Until now the plugin overwrites the `servicePath` property in Serverless during its lifecycle and
sets it back in the cleanup lifecycle event. This makes the whole system very fragile and dependent on
the consistent use of the property in Serverless as well as all other plugins. In the extreme, we can
throw the dice, if the next Serverless version will lead to problems or not.

This behavior should be changed and the plugin should use the given `servicePath` as plugins are
expected to do.

I think the change is not trivial, because there has to be a reason, why this approach has been taken.
We should analyze every single step in the plugin and check if there are any issues that can arise from
changing it to a stable use.

After the implementation has been done, it should be tested properly and ripen in master for a while.

### LINKED ISSUE #181: V3: Deploying a single function with individual packaging does not work
The whole project deploys correctly, but when deploying individual functions I'm getting the following error:

```
Time: 6122ms
                                     Asset     Size  Chunks                    Chunk Names
functions/model/job_request/get_details.js  3.84 MB       0  [emitted]  [big]  functions/model/job_request/get_details
Serverless: Packaging function: job-request-get_details...

  ENOENT: no such file or directory, open '<project-serverless>/.webpack/.serverless/job-request-get_details.zip'
```

HyperBrain:
## What did you implement:

Closes #181 
Closes #120 
Closes #173 
Closes #117 
Closes #116 
Relates to #60 

## How did you implement it:

If `package.individually: true` is set in the service, the individual entry points (functions) are now compiled by using WebPack's multi-compile feature. That means, that each function is optimized separately as well as the externals for every function. The node_modules for each function are then packaged individually, so that every resulting function zip contains only the needed modules and the individually optimized function handler.

Additionally the plugin does not change `serverless.config.servicePath` anymore, so that other plugins and the framework itself behave as expected and work during any lifecycle event and Serverless does not interfere with the packaging of the optimized node modules anymore.

### Further changes
* Added ESLint as preparation for 3.0.0, so that the implementation is conformant.
* Added unit tests for the new functionality (as well as for old ones where tests where missing).
* Added verbose output with `--verbose`

## How can we verify it:

Use a simple service that defines multiple functions, where each function uses a different set of dependencies. Use a webpack config that uses `lib.entries` for entry and set the package wide individual packaging in the service.

Now run `serverless package` and check the generated function zip files in the `.serverless` folder.

I checked it with one of our production projects, including tests with `serverless invoke local`. Found no issues. The only thing I have to verify is, that deployments without individual packaging still work.

## Todos:

- [x] Write tests
- [x] Write documentation
- ~~[ ] Enable "Allow edits from maintainers" for this PR~~


lib/packExternalModules.js
```diff
@@ -67,49 +65,87 @@ function getExternalModules(stats) {
...
-      const externalModules = getExternalModules(stats);
-
-      // this plugin will only install modules stated in dependencies section of package.json
-      const prodModules = getProdModules(externalModules, packagePath);
-
-      if (prodModules.length === 0) {
-        return;
-      }
+    if (!includes) {
+      return BbPromise.resolve();
```
johnf: should be
``` javascript
return BbPromise.resolve().return(stats);
```
or the next module errors out with null for stats.

HyperBrain: Good catch. Will change it.

johnf:
@HyperBrain What should be different if this works? More files inside the zip?

HyperBrain:
No. It will package only the node modules that are used by the function that is being packaged. Previously if you enabled individual packaging, all the packages were the same (it even packaged all functions in all function packages), i.e. it just duplicated the service package file (without the individually option). The new packaging is visible as soon as you also set "webpackIncludeModules".

The second major improvement is, that now every function is compiled individually, which leads to much better optimization, because every function is optimized separately. This will show up, when you enable tree shaking in WebPack.

And last, it prepares to use manual multi-compiles. The next step will be, that you can emit an array of webpack configurations from your webpack.conf, that will automatically trigger the multi-compiler. The new packaging infrastructure will allow for that without having to add any new implementation - only validate has to be touched to make it work.

johnf:
nm. I had wrong config in serverless.yml. I can now see a different zip file per function.

Some of them look wright and have a single js. Others seem to be 8M in size and have the while source inside.

I'll have more time on the weekend to try and work out what the difference is

HyperBrain:
You can also try to use "lib.entries" for the webpack entry definition that is available since 2.2.0.
This makes sure that the plugin sets the correct entry points.
```
// webpack.conf
module.exports= {
  ...
  entry: slsw.lib.entries,
  ...
  output:
    ...
    filename: '[name].js',
    ...
};
```
This makes the use of the plugin quite easy and eliminates the chances for errors.

HyperBrain:
Switched one of our production servcies to this branch now. After the latest fixes it packages everything correctly and independently and even works with git dependencies and the serverless-warmup plugin (which creates a separate function dynamically during the package process).

Also works with separate build and deployment (using the `--package` option) from different servers.

cwaltken-edrans:
On this branch I get the following error... Adding underscore as a dependency did not solve the issue...

```
  Reference Error ----------------------------------------
 
  _ is not defined
 
     For debugging logs, run again after setting the "SLS_DEBUG=*" environment variable.
 
     Forums:        forum.serverless.com
     Chat:          gitter.im/serverless/serverless
 
     OS:                     linux
     Node Version:           6.11.1
     Serverless Version:     1.18.0

```

HyperBrain:
Fixed in a minute... I checked in the wrong file ;-)

HyperBrain:
@cwaltken-edrans Can you update and try again?

cwaltken-edrans:
That fixes the lodash error. Still I'm back at this one:
```
  Type Error ---------------------------------------------
 
  lambda is not a function
 
     For debugging logs, run again after setting the "SLS_DEBUG=*" environment variable.
 
     Forums:        forum.serverless.com
     Chat:          gitter.im/serverless/serverless
 
     OS:                     linux
     Node Version:           6.11.1
     Serverless Version:     1.18.0
 ```

HyperBrain:
I used the invoke local multiple times today and had no problems so far. It might be a specific configuration in your project. The best way to investigate it would be to have a quite simple test project that reproduces the error.
Can you create an issue for the problem? Then we can handle it there and we could put all information we have into it.

cwaltken-edrans:
> I used the invoke local multiple times today and had no problems so far. It might be a specific configuration in your project. The best way to investigate it would be to have a quite simple test project that reproduces the error.
Can you create an issue for the problem? Then we can handle it there and we could put all information we have into it.

I've just checked again. I broke the entry function trying to fix the issues I had previously. I reverted the changes and now i can build my function.

Sadly my initial issue is still there. Functions invoked locally don't have access to files bundled with the file-loader it seems. I'll open an issue on that.

HyperBrain:
Yes, please do that. I did not do in-depth tests yet with different loaders, so there might be still issues in there that have to be fixed. A big thank you for taking the effort to test 🥇

PR #234, opened 2017-09-21
Log errors in webpack config explicitly.

HyperBrain:
## What did you implement:

This PR adds an extra log line if the webpack config file could not be loaded.
This makes the reason for errors in the configuration more clear than just printing the
actual exception without context.

## How did you implement it:

Added a try/catch around the `require()` of the configuration file.

## How can we verify it:

Implement a syntax error in your webpack config. Any try to package or deploy should mention
to the configuration file now before printing the error message or stacktrace.

## Todos:

- [x] Write tests
- [ ] Write documentation
- [x] Fix linting errors


HyperBrain:
Released with `3.1.0`

PR #237, opened 2017-09-22
Packaging in parallel instead of series

### LINKED ISSUE #236: Packing external module time
# This is a Feature Proposal

## Description

I'm using serverless webpack with about a dozen of lambda.
When I run `serverless deploy`, it is taking about 10 minutes to deploy my service, and the most part of it is actually the `Packing external module` step.
I just added `--verbose option` and it seems that every functions are packaged one after another.
Is there a reason to not run in parallel ? I think this could be changed here: https://github.com/serverless-heaven/serverless-webpack/blob/master/lib/packExternalModules.js#L213

I could do a PR if necessary, this could really save us a lot of time.

## Additional Data

* ***Serverless-Webpack Version you're using***: 3.0.0
* ***Webpack version you're using***: 3.3.0
* ***Serverless Framework Version you're using***: 1.18.1
* ***Operating System***: macOS

ChristopheBougere:
## What did you implement:

Closes #236 

The goal is to execute in parallel the external modules packaging instead of doing it in series. This can increase the speed of the `serverless deploy` a lot if you have many functions.

## How did you implement it:

Just changing a array promise from series to parallel.

## How can we verify it:

Test still pass, and deploy still works on my side.

## Todos:

- [ ] Write tests
- [ ] Write documentation
- [ ] Fix linting errors

### DIFF
--- a/lib/packExternalModules.js
+++ b/lib/packExternalModules.js
@@ -249,6 +249,8 @@ module.exports = {
           })
           .tap(() => this.options.verbose && this.serverless.cli.log(`Prune: ${modulePath} [${_.now() - startPrune} ms]`));
         });
+      }, {
+        concurrency: 5
       })
       .return(stats);
     });
--- a/lib/packExternalModules.js
+++ b/lib/packExternalModules.js
@@ -210,7 +210,7 @@ module.exports = {
           reject(e);
         });
       })
-      .mapSeries(compileStats => {
+      .map(compileStats => {
         const modulePath = compileStats.compilation.compiler.outputPath;
 
         // Create package.json


HyperBrain:
@ChristopheBougere Thanks for the PR 👍  As the code changes are trivial, they should not break any unit tests. However, I will do some integrative tests with one of our projects that contains lots of functions too and see if it works there. If that succeeds, I'll merge it and it will be part of version `3.1.1`.

HyperBrain:
@ChristopheBougere I tested this with one of our projects and it worked (although the timing outputs of the copies merely all show the summarized time now, because the all start at the same time).

There's one thing that we still might change. Imo it would be good, to have a limitation of the concurrently running copies because now it is unlimited and a service with 100 Lambdas will do 100 at the same time with 100x memory needed for one.
I think we can limit it to 5 or 10 by using `map`'s concurrency option, just to be on the safe side.

HyperBrain:
After a further test I'm sure that limiting to 5 is more appropriate than 10, because it spawns as many prune processes as you have Lambdas in parallel, which can eat up your memory quite fast 😃

ChristopheBougere:
Thanks for the feedback @HyperBrain, that make sense!
I'll commit this, let me know if I should add the limit as a parameter.

HyperBrain:
I'll give it a try with the same project later today or tomorrow.

HyperBrain:
I'm out for some days. Will test it once more end of this week.

wenderjean:
@HyperBrain @ChristopheBougere I'm getting these such errors when trying to `package` using the solution:

`node_modules/.bin/serverless package --verbose`

## NPM V3
```

  Command failed: npm prune
  npm WARN package.json @ No description
  npm WARN package.json @ No repository field.
  npm WARN package.json @ No README data
  npm WARN package.json @ No license field.
  npm ERR! cb() never called!


     For debugging logs, run again after setting the "SLS_DEBUG=*" environment variable.

     Forums:        forum.serverless.com
     Chat:          gitter.im/serverless/serverless

     OS:                     darwin
     Node Version:           6.10.0
     Serverless Version:     1.23.0
```

## NPM V5
```
 spawn /bin/sh EMFILE
```

HyperBrain:
@wenderjean Thanks for the test. This shows that, while the copy might be parallelized, the `npm prune` might not - it makes excessive use of the npm cache which might exceed the OS open files limit.
I'm not sure how to proceed here. At least the tests showed that the implementation is heavily dependent on the operating system configuration. Maybe a better approach would be to limit it to 3 by default, but make the whole parallelization optional (with a switch). Any thoughts?

arabold:
The cleanest would seem to copy files in parallel (as that's a simple file system operation which should benefit from parallelism due to many small files), but run all `npm` calls afterwards in a serialized way.

wenderjean:
I'll give it a try @arabold, I'm not sure about how much the time will improve because the current benchmark, at least in my hardware is:

```
npm install = 40 ~ 50 seconds
copy = 3 ~ 4 seconds each function
prune = 3 ~ 4 seconds each function
```
We're gonna probably gain 1 or 2 improvement seconds.

arabold:
I might find some time next week (I'm out this one) to test this with our setup as well. I'm curious what the actual numbers are.

ChristopheBougere:
Sorry for the delay.
@wenderjean That's interesting. I haven't seen that since I'm using it (NPM 5), but yes it may depends on the OS, and computer config...

So as you've said, this may not be the best way to improve performances...
Another idea: could we keep the `node_modules` folder after having deployed (would imply not deleting the `.webpack` folder), and if for the next deployment dependencies are identical we could keep them?
This would be a bigger change, but it might really improve performances

HyperBrain:
@ChristopheBougere Interesting thought 👍 . As the temporary packaging directory structure is:
```
root
  +-- .webpack
      +-- dependencies
      +-- service (for service packaging)
      +-- xxx (for individual packaging)
```
and the composite dependencies (all dependencies) are installed first to `dependencies` it would be an option to keep `.webpack/dependencies` but remove anything else from the webpack folder on cleanup. Keeping the compiled out put might lead to remnants in the compiled directories that would be packaged and pollute the packages.

On a subsequent build the install of the dependencies would be much faster as there are most likely only minor changes in the dependencies. Of course this would only speed up the initial install time for now.

ChristopheBougere:
Closing this PR as this solution may have side effects.

PR #246, opened 2017-10-02
Use package-lock.json to install packages if there is one available.

### LINKED ISSUE #245: package-lock.json is not respected when bundling
# This is a Bug Report

## Description

When bundling a project the `serverless-webpack-plugin` doesn't respect the `package-lock.json` which leads to involuntary updates of the dependencies.

## Expected Behavior

The plugin must respect the `package-lock.json` if exists and never upgrade components on its own.

HyperBrain:
## What did you implement:

Closes #245

## How did you implement it:

If a `package-lock.json` is found besides the `package.json` of the project, it is copied to the packaged dependencies folder, so that the plugin now installs the locked versions (for components that are present in the lock file).
The `package.json` used for packaging is now pre-assembled because only an unqualified `npm install` uses an existing lock file.

## How can we verify it:

Use a project with locked dependencies, package and check the created zip file.

## Todos:

- [x] Write tests
- [ ] Write documentation
- [x] Fix linting errors


lib/packExternalModules.js
```diff
@@ -193,22 +192,49 @@ module.exports = {
...
+        }
+        const moduleVersion = _.join(_.tail(splitModule), '@');
+        _.set(compositePackage, `dependencies.${_.first(splitModule)}`, moduleVersion);
+      });
+      this.serverless.utils.writeFileSync(compositePackageJson, JSON.stringify(compositePackage, null, 2));
 
-      return new BbPromise((resolve, reject) => {
+      // (1.a.2) Copy package-lock.json if it exists, to prevent unwanted upgrades
+      const packageLockPath = path.join(path.dirname(packageJsonPath), 'package-lock.json');
+      return BbPromise.fromCallback(cb => fse.pathExists(packageLockPath, cb))
```
arabold: Why not use the sync versions if you're using it in a synchronous way?

HyperBrain: Because it will stall the Node event loop (even if something else might be running in the background - outside of the function) - we still run in the Serverless context and should not affect our host in any way ;-)

arabold: Tell that to the Serverless guys using synchronous functions whenever humanly possible 🤣

HyperBrain: Haha... In the meantime they even turned some of the stuff asynchronously, like file copy.

lib/packExternalModules.js
```diff
@@ -193,22 +192,49 @@ module.exports = {
...
+      };
+      _.forEach(compositeModules, compositeModule => {
+        const splitModule = _.split(compositeModule, '@');
+        // If we have a scoped module we have to re-add the @
+        if (_.startsWith(compositeModule, '@')) {
+          splitModule.splice(0, 1);
+          splitModule[0] = '@' + splitModule[0];
+        }
+        const moduleVersion = _.join(_.tail(splitModule), '@');
+        _.set(compositePackage, `dependencies.${_.first(splitModule)}`, moduleVersion);
```
HyperBrain: @silver2k This is the issue. The assembly of the dependency name breaks here if the dependency includes a `.`.

It is handled as if it was a nested property. I will provide a fix asap and release 3.1.2.

PR #256, opened 2017-10-19
Support skipping webpack on a per function basis

### LINKED ISSUE #255: Skip webpack on a per function basis
# This is a Feature Proposal

## Description

For example

```yaml
functions:
  pthon_func:
    runtime: python2.7
    webpack: false
```

johnf:
## What did you implement:

Closes #255

## How did you implement it:

We add ```webpack: false``` to functions where we want to skip webpack.
In ```index.js``` in the relevant hooks we check if this is set and just return. This will run the original hook.

## How can we verify it:

Examples:
In serverless.yml add a function similar to
```yaml
functions:
  get_oai:
    runtime: python2.7
    handler: src/oai.handler
    webpack: false
```

run ```sls deploy -f get_oai``` and the .serverless/get_oai.zip should contain everything in the project. Webpack will not run

## Todos:

- [x] Write tests
- [x] Write documentation
- [x] Fix linting errors

### DIFF
--- a/lib/validate.js
+++ b/lib/validate.js
@@ -6,6 +6,7 @@ const fse = require('fs-extra');
 const glob = require('glob');
 const lib = require('./index');
 const _ = require('lodash');
+const helpers = require('./helpers');
 
 /**
  * For automatic entry detection we sort the found files to solve ambiguities.
@@ -56,7 +57,6 @@ module.exports = {
       }
       return path.extname(_.first(sortedFiles));
     };
-
     const getEntryForFunction = (name, serverlessFunction) => {
       const handler = serverlessFunction.handler;
 
@@ -84,14 +84,21 @@ module.exports = {
     const entries = {};
 
     const functions = this.serverless.service.getAllFunctions();
+
     if (this.options.function) {
-      const serverlessFunction = this.serverless.service.getFunction(this.options.function);
-      const entry = getEntryForFunction.call(this, this.options.function, serverlessFunction);
-      _.merge(entries, entry);
-    } else {
-      _.forEach(functions, (func, index) => {
-        const entry = getEntryForFunction.call(this, functions[index], this.serverless.service.getFunction(func));
+      const functionName = this.options.function;
+      if (helpers.hasNodeRuntime(functionName, this.serverless)) {
+        const serverlessFunction = this.serverless.service.getFunction(functionName);
+        const entry = getEntryForFunction.call(this, this.options.function, serverlessFunction);
         _.merge(entries, entry);
+      }
+    } else {
+      _.forEach(functions, (functionName, index) =>  {
+        if (helpers.hasNodeRuntime(functionName, this.serverless)) {
+          const serverlessFunction = this.serverless.service.getFunction(functionName);
+          const entry = getEntryForFunction.call(this, functions[index], serverlessFunction);
+          _.merge(entries, entry);
+        }
       });
     }
 
--- a/README.md
+++ b/README.md
@@ -108,6 +108,20 @@ module.exports = {
 };
 ```
 
+#### Non JS functions
+
+You can mix and match a project with different runtimes and serverless-webpack
+will do the right thing. However, you must package individually for this to
+work.
+
+```yaml
+# serverless.yml
+...
+package:
+  individually: true
+...
+```
+
 #### Full customization (for experts)
 
 The lib export also provides the `serverless` and `options` properties, through
--- a/lib/compile.js
+++ b/lib/compile.js
@@ -3,9 +3,14 @@
 const _ = require('lodash');
 const BbPromise = require('bluebird');
 const webpack = require('webpack');
+const helpers = require('./helpers');
 
 module.exports = {
   compile() {
+    if (helpers.hasEmptyWebpackConfig(this.webpackConfig)) {
+      return BbPromise.resolve();
+    }
+
     this.serverless.cli.log('Bundling with Webpack...');
 
     const compiler = webpack(this.webpackConfig);
--- a/lib/helpers.js
+++ b/lib/helpers.js
@@ -1,0 +1,19 @@
+'use strict';
+
+const _ = require('lodash');
+
+function hasNodeRuntime(funcName, serverless) {
+  const serverlessFunction = serverless.service.getFunction(funcName);
+  const runtime = serverlessFunction.runtime || serverless.service.provider.runtime;
+
+  return !runtime || /^nodejs/.test(runtime);
+}
+
+function hasEmptyWebpackConfig(config) {
+  return (_.isEmpty(config) || (!_.isArray(config) && _.isEmpty(config.entry)));
+}
+
+module.exports = {
+  hasNodeRuntime,
+  hasEmptyWebpackConfig,
+};
--- a/lib/packExternalModules.js
+++ b/lib/packExternalModules.js
@@ -6,6 +6,7 @@ const path = require('path');
 const childProcess = require('child_process');
 const fse = require('fs-extra');
 const isBuiltinModule = require('is-builtin-module');
+const helpers = require('./helpers');
 
 /**
  * Add the given modules to a package json's dependencies.
@@ -159,6 +160,9 @@ module.exports = {
    * and performance.
    */
   packExternalModules() {
+    if (helpers.hasEmptyWebpackConfig(this.webpackConfig)) {
+      return BbPromise.resolve();
+    }
 
     const stats = this.compileStats;
 
--- a/lib/packageModules.js
+++ b/lib/packageModules.js
@@ -7,6 +7,7 @@ const archiver = require('archiver');
 const fs = require('fs');
 const glob = require('glob');
 const semver = require('semver');
+const helpers = require('./helpers');
 
 function setArtifactPath(funcName, func, artifactPath) {
   const version = this.serverless.getVersion();
@@ -78,6 +79,10 @@ function zip(directory, name) {
 
 module.exports = {
   packageModules() {
+    if (helpers.hasEmptyWebpackConfig(this.webpackConfig)) {
+      return BbPromise.resolve();
+    }
+
     const stats = this.compileStats;
 
     return BbPromise.mapSeries(stats.stats, (compileStats, index) => {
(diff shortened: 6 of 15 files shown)


HyperBrain:
@johnf Thanks for the PR 👍 .

One question: Normally all functions with runtimes other than node should not be webpacked in a mixed project. So woudn't it be better to disable webpack for all these runtimes rather than having the user to add a flag where the plugin actually knows that it would not work?
Having this duplicated and force the user manually to add the flags to get the project compile seems a bit cumbersome to me.

index.js
```diff
@@ -90,24 +90,42 @@ class ServerlessWebpack {
...
         .then(this.cleanup),
 
-      'before:deploy:function:packageFunction': () => BbPromise.bind(this)
-        .then(this.validate)
-        .then(this.compile)
-        .then(this.packExternalModules)
-        .then(this.packageModules),
+      'before:deploy:function:packageFunction': () => {
+        if (this.options.functionObj.webpack === false) {
+          return;
```
HyperBrain: All hooks must return promises. We cannot guarantee that the caller is aware that a non-promise can be returned that cannot be awaited. As we are a plugin we must make no assumptions about our host.
```
return BbPromise.resolve();
```

index.js
```diff
@@ -126,15 +144,27 @@ class ServerlessWebpack {
       'webpack:serve:serve': () => BbPromise.bind(this)
         .then(() => BbPromise.reject(new this.serverless.classes.Error('serve has been removed. Use serverless-offline instead.'))),
 
-      'before:offline:start': () => BbPromise.bind(this)
-        .then(this.prepareOfflineInvoke)
-        .then(this.compile)
-        .then(this.wpwatch),
+      'before:offline:start': () => {
+        if (this.options.functionObj.webpack === false) {
```
HyperBrain: Offline is used to simulate the service, not a single function (and will recompile in case of a file changed).
So imo the checks on `offline:start` and `offline:start:init` must be removed, as this would be never called.

johnf:
@HyperBrain That is an excellent point. I was so focussed on making my problem go away I didn't think of that!
I'll update with that change this weekend.

lib/validate.js
```diff
@@ -83,6 +83,10 @@ module.exports = {
       _.merge(entries, entry);
     } else {
       _.forEach(functions, (func, index) => {
+        const data = this.serverless.service.getFunction(func);
+        if (data.webpack === false) {
```
HyperBrain: Check for the runtime here - not the manually set flag, as discussed in my comment.
Additionally you create an entries array here that is shorter than the actual number of functions in the service.
This might fail in the compile and deploy steps with individual packaging when the entries have to be mapped back to the actual functions by index. A better way would be to only iterate over the supported functions (i.e. node.js) in all places then, instead of doing the check within the for loop. That would catch all possible combinations.

```
const supportedFunctions = _.filter(functions, func => !func.runtime || _.includes([ 'nodejs6.10', ... ] func.runtime);
_.forEach(supportedFunctions, (func, index) =>  {
```

HyperBrain:
@johnf We also should check that the changes do not affect other providers like Google and OpenWhisk that currently work with the plugin. Not sure yet, how to do that - maybe we have to to some tests with the PR there.

johnf:
@HyperBrain 

So If we want to be 100% correct and support all the providers then I've come up with

```javascript
    this.nodeRuntime = () => {
      const provider = this.serverless.service.provider.name;
      const runtime = this.options.functionObj.runtime || this.serverless.service.provider.runtime;
      
      const jsOnlyProviders = [ 'azure', 'google', 'webtasks' ];
      if (jsOnlyProviders.include(provider)) {
        return true;
      }
      
      if (provider === 'openwhisk' && runtime === null) {
        return true;
      }

      return /^nodejs/.test(runtime);
    };
```

I think this might be a bit tedious and will be prone to errors if things change in the future.

I think instead we assume that the top level runtime is a javascript runtime and then we just check if a specific provider has been set on the function. (I could document this in the readme).

```javascript
this.nodeRuntime = () => {
      const runtime = this.options.functionObj.runtime;
      
      return !runtime || /^nodejs/.test(runtime);
    };
```

Thoughts?

lib/validate.js
```diff
@@ -77,14 +84,20 @@ module.exports = {
...
     const functions = this.serverless.service.getAllFunctions();
+
     if (this.options.function) {
       const serverlessFunction = this.serverless.service.getFunction(this.options.function);
-      const entry = getEntryForFunction.call(this, this.options.function, serverlessFunction);
-      _.merge(entries, entry);
-    } else {
-      _.forEach(functions, (func, index) => {
-        const entry = getEntryForFunction.call(this, functions[index], this.serverless.service.getFunction(func));
+      if (this.hasNodeRuntime(serverlessFunction)) {
```
HyperBrain: @johnf Did you check if this works with `serverless invoke local` in case you invoke a non-node function?
The entries object will end up empty in this case (because the single function is skipped - which is ok), but what will webpack do in this case?

johnf: @HyperBrain 

I've done a bunch of testing of different scenarios. Amended the code to skip running webpack in the right places.

One caveat though, in a mixed project you have to package individually. This is because if you just do an ```sls deploy``` the serverless-webpack code takes over and only packages up the js files. It will ignore the rest of the project.

I've added a note to the README. Feel free to edit or move it somewhere else in there.

johnf:
except I've broken the tests because of the way I added in ```hasNodeRuntime```. Will review that a bit later have to head out

johnf:
@HyperBrain Any thoughts on preferred approach here.
I added some common helper functions into index.js which works at runtime because the constructor runs, but doesn't work in the unit tests.
I could either put the helpers into a helpers.js and include them in every file or maybe just pull them in, in the tests themselves?

HyperBrain:
Imo, the helper functions should be put into a separate file that exports them (to keep the same semantics throughout the project). I have one question to the empty config check - I don't completely understand the reason for that. Wouldn't it mean in this case, that there are no node functions at all in the project, or it is misconfigured? Is having a project with the plugin enabled but only Python methods defined valid? Can you please shed come ligt onto that?

johnf:
@HyperBrain 
I've split out the helper functions into ```helpers.js```

With regards to the empty check it depends on the call path. For things like packModules an earlier hook will have gone through validate but I don't think I was able to pull the runtime info anymore. So it was easier to check if a webpack config had been generated and skip on that basis. It usually happens when something is called with ```-f python_module```.

HyperBrain:
Looks good now. I'll do some tests with our projects and fix the unit tests to test for the newly introduced code branches, so that Coveralls is happy again.

mjmac:
Hi, is there anything I can do to help get this integrated? Just ran into this issue today and was excited to find that there might already be a solution, but then heard the sad trombone when I saw that it's not landed yet...

HyperBrain:
@mjmac Oh sorry, I lost this entirely from my sight. If someone could add the unit tests, so that code coverage stays up again, I'll merge it (just do a PR on top of this one).

OrKoN:
I have tried to add test but found that it's not easy because of mutations and mocks. Is there another way to skip Webpack for certain functions, perhaps, in https://github.com/serverless-heaven/serverless-webpack/blob/master/index.js so that the lib does not need to be changed a lot?

OrKoN:
I have tried to do something myself over here https://github.com/serverless-heaven/serverless-webpack/compare/master...OrKoN:skip-non-node-functions?expand=1 It seems to work but the non-node functions get packaged weirdly: it looks like every file from the project is added to the archive and include/exclude settings are ignored.

```
  test:
    runtime: go1.x
    handler: go/bin/main
    package:
      individually: true
      exclude:
        - ./**
      include:
        - go/bin/main
```

HyperBrain:
@OrKoN Sorry for the huge delay. I think the issue with packaging everything is, that with the conditional package property on the functions, Serverless' own packaging is triggered when it inspects these functions.

This leads to the fact that Serverless starts its complete packaging for the service and just overwrites webpacks package output (and Serverless will just include _everything_).

I think we must be a bit more experimental and sophisticated here and _tune_ the in-memory service definition in a way that we hide any triggering parts from serverless or just call the very specific Serverless function packaging from our own packager, instead of having it act as fallback (and thus destroying all our outputs)

OrKoN:
@HyperBrain no problems. It seems like it's a very hacky workaround :-) For now, I decided that I am not going to mix different languages in one project. But the feature seems to be very useful, eventually.

liam-betsworth:
@HyperBrain is there any progress on this?

I'm also coming up against this problem using a serverless configuration with multiple environments.

This is a real deal breaker

HyperBrain:
Hi @liam-betsworth , I agree that this issue gets more pressing as more runtimes appear (at least on AWS). As this implementation is somehow hacky, I'm still thinking about a proper solution.
The main solution would be to mark and remember all non-JS functions while the plugin iterates the service functions. Then it should skip the modification of the function.package.artifact property for the found functions. That will lead to a Serverless packaging for these non-JS functions (with all its drawbacks). However this approach would only work with individual packaging, but I think that should be ok.

20112077:
Are there any updates on this? I am also trying to mix nodejs and python functions in one serverless project.

gdugernier:
> Since this PR hasn't been merged yet; Do you have suggestions on forcing this behavior without changing the plugin? Is there a way to use webpack.config.js to skip over functions?

Facing the same issue. I have both python2.7 and JS functions in my project. I'm currently I'm using [lerna](https://github.com/lerna/lerna) and have 2 packages: one with the JS functions and serverless-webpack, and one with the python without serverless-webpack. It's somewhat hacky and necessitate some cross-packages ${cf:}, but it's doing the trick.

I'm not sure it's the most elegant way but since I was already using Lerna to separate lambda resources and database/vpc resources it felt the easiest way to implement

gparonitti:
Any update on this? What does it need to get integrated? This feature would be of much help 😄 .

jagregory:
Any update on this?

I'm currently maintaining a forked version of serverless-webpack for this functionality. It's not ideal.

ineffyble:
Any updates on this? Very desired functionality.

stlouisweb:
@HyperBrain can someone review https://github.com/serverless-heaven/serverless-webpack/pull/530? It appears to be a competing for this functionality.

j0k3r:
See https://github.com/serverless-heaven/serverless-webpack/pull/579 available since 5.3.3

nponeccop:
I've almost started working on resolving the conflicts yesterday lol.

@j0k3r you can use me as a human merge conflict resolution bot. Which PR should I rebase so you merge it and I work on next one?

PR #261, opened 2017-10-26
Allow webpack stats option to overwrite default behaviour

### LINKED ISSUE #260: Use stats option to overwrite console.log behaviour
# This is a Feature Proposal
## Description
Currently, the stats options in webpack.config.js seems to be ignored. No matter what I put in my webpack.config.js, the console output is the same, which I find them too verbose and annoying, especially when developing & watch. E.g: 

image

I think it would be good if we can allow webpack stats config to overwrite the default behaviour.

[Update] 
After digging through the source, I think the we can do it easily by updating [lib/compile.js](https://github.com/serverless-heaven/serverless-webpack/blob/5c9f14d53f4ce739ebcf69748d532c0f5253b7c3/lib/compile.js#L24)
and [lib/wpwatch.js](https://github.com/serverless-heaven/serverless-webpack/blob/5c9f14d53f4ce739ebcf69748d532c0f5253b7c3/lib/wpwatch.js#L25)

I can make a PR if you're ok with this.

* Can't find any

## Additional Data

* ***Serverless-Webpack Version you're using***: 3.1.2
* ***Webpack version you're using***: 3.8.1
* ***Serverless Framework Version you're using***: 1.23.0

dangcuuson:
## What did you implement:

Closes #260

## How did you implement it:

When console log the compile stats, if webpack.config.js has stats configured, use it. Otherwise use default behaviour.

## How can we verify it:

Using example/serverless-offline with default config:

image

After add `stats: 'minimal'` to webpack config:
image

## Todos:

- [ ] Write tests
- [ ] Write documentation
- [ ] Fix linting errors

### DIFF
--- a/README.md
+++ b/README.md
@@ -144,6 +144,24 @@ module.exports = {
 };
 ```
 
+### Stats
+
+By default, the plugin will print a quite verbose bundle information to your console. However, if 
+you are not satisfy with the current output info, you can overwrite it in your `webpack.config.js`
+
+```js
+// webpack.config.js
+
+module.export = {
+  // ...
+  stats: 'minimal'
+  // ...
+}
+```
+
+All the stats config can be found in [webpack's documentation](https://webpack.js.org/configuration/stats/)
+
+
 ### Node modules / externals
 
 By default, the plugin will try to bundle all dependencies. However, you don't
--- a/lib/compile.js
+++ b/lib/compile.js
@@ -19,15 +19,16 @@ module.exports = {
         }
 
         const compileOutputPaths = [];
+        const consoleStats = this.webpackConfig.stats || {
+          colors: true,
+          hash: false,
+          version: false,
+          chunks: false,
+          children: false
+        };
 
         _.forEach(stats.stats, compileStats => {
-          this.serverless.cli.consoleLog(compileStats.toString({
-            colors: true,
-            hash: false,
-            version: false,
-            chunks: false,
-            children: false
-          }));
+          this.serverless.cli.consoleLog(compileStats.toString(consoleStats));
 
           if (compileStats.compilation.errors.length) {
             throw new Error('Webpack compilation error, see above');
--- a/lib/wpwatch.js
+++ b/lib/wpwatch.js
@@ -22,7 +22,7 @@ module.exports = {
       }
 
       if (stats) {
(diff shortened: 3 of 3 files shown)


HyperBrain:
Thanks. The implementation looks good 🙌 
Could you additionally add something to the README that mentions the possibility to override the stats with a proper webpack config (maybe only that it is supported)? Then poeple do not have to figure it out by asking questions.

dangcuuson:
@HyperBrain it's done. :smile:

dashmug:
This doesn't do anything on my project.

Here's my config:
```
'use strict'

const path = require('path')
const { CheckerPlugin } = require('awesome-typescript-loader')
const slsWebpack = require('serverless-webpack')
const webpack = require('webpack')

module.exports = {
    entry: slsWebpack.lib.entries,
    target: 'node',
    output: {
        libraryTarget: 'commonjs2',
        path: path.join(__dirname, 'build'),
        filename: '[name].js',
    },
    module: {
        loaders: [
            { test: /\.ts$/, loader: 'awesome-typescript-loader' },
            { test: /\.ejs$/, loader: 'ejs-loader' },
        ],
    },
    externals: ['bcrypt', 'aws-sdk', 'aws-xray-sdk'],
    resolve: {
        extensions: ['.ts', '.js'],
    },
    devtool: 'source-map',
    plugins: [
        new CheckerPlugin(),
        new webpack.DefinePlugin({
            'process.env.NODE_ENV': JSON.stringify('production'),
        }),
    ],
    stats: 'minimal',
}
```

Package versions:
```
    serverless: ^1.24.1
    serverless-webpack: ^4.0.0
    webpack: ^3.8.1
```

HyperBrain:
@dashmug That's strange. 
@dangcuuson Any idea?

dangcuuson:
@HyperBrain not sure. I'll have a look after work.
@dashmug could you tell me which command you use to run your serverless? Did you use serverless-offline, or local invoke?

dashmug:
I used `sls webpack`.

dangcuuson:
I found the problem:

If you have `package` option in your `serverless.yml`, it will [map your webpack config into an array](https://github.com/serverless-heaven/serverless-webpack/blob/master/lib/validate.js#L184), which kind of break my implementation.

The fix should be easy.

PR #264, opened 2017-10-30
Exclude node_modules completely if deploying to google cloud functions, as GCF installs package.json.

debanjanbasu:
## Todos:

- [x] Write tests
- [ ] Write documentation
- [x] Fix linting errors

### DIFF
--- a/lib/packExternalModules.js
+++ b/lib/packExternalModules.js
@@ -292,19 +292,19 @@ module.exports = {
         this.serverless.utils.writeFileSync(modulePackageJson, JSON.stringify(modulePackage, null, 2));
 
         // Copy modules
-        const startCopy = _.now();
-        return BbPromise.fromCallback(callback => fse.copy(path.join(compositeModulePath, 'node_modules'), path.join(modulePath, 'node_modules'), callback))
-        .tap(() => this.options.verbose && this.serverless.cli.log(`Copy modules: ${modulePath} [${_.now() - startCopy} ms]`))
-        .then(() => {
-          // Prune extraneous packages - removes not needed ones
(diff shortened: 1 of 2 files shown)


HyperBrain:
Hi @debanjanbasu , thank you for the PR. Can you fill out the description template completely - especially the checkboxes and check "enable edits from maintainers" in your PR?

lib/packExternalModules.js
```diff
@@ -291,20 +291,23 @@ module.exports = {
...
-          return BbPromise.fromCallback(callback => {
-            childProcess.exec('npm prune', {
-              cwd: modulePath
-            }, callback);
-          })
-          .tap(() => this.options.verbose && this.serverless.cli.log(`Prune: ${modulePath} [${_.now() - startPrune} ms]`));
-        });
+        // Copy modules only if not google-cloud-functions
+        // GCF Auto installs the package json
+        if (_.get(this.serverless, 'service.provider.name') !== 'google') {
```
HyperBrain: I would prefer an early exit here if the provider is google instead of putting the default case into the if.
```
if (_.get(this.serverless, 'service.provider.name') === 'google') {
  return BbPromise.resolve();
}
...
```

HyperBrain: Updated this. I will add unit tests for the new behavior to restore coverage.

HyperBrain:
@debanjanbasu May I ask you for a favor? There is still the column for GCF missing in the provider compatibility chart (README). Would you mind to prepare a PR that adds the GCF status to the README?
I do not use Google, so I'm not able to check whether it works or not 😄

PR #265, opened 2017-10-30
Update README.md to fix sample config.

thepont:
Path needs to be an absolute path since webpack 2.3.1 current sample code in documentation will not run.

`- configuration.output.path: The provided value ".webpack/service" is not an absolute path!`

## What did you implement:
Fix for a sample in the documentation, webpack requires `path` to be absolute.

Closes #XXXXX

## How did you implement it:

Starting with the sample code, I adjusted it so it worked. using path.resolve to resolve the abs path for web pack.

## How can we verify it:

Check that sample correctly runs by running the webpack.config.js

## Todos:

- [ ] Write tests
- [X] Write documentation
- [ ] Fix linting errors

### DIFF
--- a/README.md
+++ b/README.md
@@ -132,12 +132,13 @@ for best compatibility with external dependencies:
 
 ```js
 // webpack.config.js
+const path = require('path');
 
 module.exports = {
   // ...
   output: {
     libraryTarget: 'commonjs',
-    path: '.webpack',
+    path: path.resolve(__dirname, '.webpack'),
     filename: 'handler.js', // this should match the first part of function handler in `serverless.yml`
   },
   // ...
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -132,12 +132,13 @@ for best compatibility with external dependencies:
...
 // webpack.config.js
+const path = require('path');
 
 module.exports = {
   // ...
   output: {
     libraryTarget: 'commonjs',
-    path: '.webpack',
+    path: path.resolve(__dirname, '.webpack'),
     filename: 'handler.js', // this should match the first part of function handler in `serverless.yml`
```
HyperBrain: Since 3.x.x the filename should also be defined as `[name].js` so that it can be used with automatic entry resolution, which is now the preferred way to set the entries.

HyperBrain: Changed it right now - ready to merge 😄

Issue #266, opened 2017-11-01
Webpack object entry with Array<string> throws error in v3


ceilfors:
# This is a Bug Report

## Description

For bug reports:
* What went wrong?
Upgrading serverless-webpack from v2 from v3 throws an error from a perfectly valid webpack.config.js.

This is happening because Instead of using the typical `entry: slsw.lib.entries` in webpack.config.js, I have changed the entry to `entry: addDynamicRequire(slsw.lib.entries)`. In a nutshell, I'm adding 'babel-polyfill', 'source-map-support/register' in webpack.config.js to avoid the need of adding these two requires in all of my handlers.

* What did you expect should have happened?
- serverless-webpack should not restrict the contract of webpack.config.js.
- or make it explicit in the documentation

* What was the config you used?
serverless.yml
package individually: true

* What stacktrace or error message from your provider did you see?
- Not sure what this means

Similar or dependent issue(s):
* None

## Additional Data

* More references:
https://webpack.js.org/concepts/entry-points/
https://stackoverflow.com/a/33396110/2464295

* ***Serverless-Webpack Version you're using***:
3.1.2

* ***Webpack version you're using***:
3.5.5

* ***Serverless Framework Version you're using***:
1.23.0

* ***Operating System***:
Mac OS

* ***Stack Trace (if available)***:
```
Type error:

  Type Error ---------------------------------------------

  Path must be a string. Received [ 'babel-polyfill',
  'source-map-support/register',
  './packages/core/src/alexa-skill.js' ]

     For debugging logs, run again after setting the "SLS_DEBUG=*" environment variable.


Stack trace:
  Stack Trace --------------------------------------------

TypeError: Path must be a string. Received [ 'babel-polyfill',
  'source-map-support/register',
  './packages/core/src/alexa-skill.js' ]
    at assertPath (path.js:28:11)
    at Object.relative (path.js:1248:5)
    at entryFunctions._.flatMap (/Users/ceilfors/github/realtime-ww2-alexa/node_modules/serverless-webpack/lib/validate.js:154:28)
    at /Users/ceilfors/github/realtime-ww2-alexa/node_modules/lodash/lodash.js:3554:27
    at /Users/ceilfors/github/realtime-ww2-alexa/node_modules/lodash/lodash.js:4944:15
    at baseForOwn (/Users/ceilfors/github/realtime-ww2-alexa/node_modules/lodash/lodash.js:3001:24)
[...]

PR #268, opened 2017-11-05
Add missing unit tests which cover package individually scenario.

ceilfors:
## What did you implement:

When trying to help on #266, I discovered that the plugin has 0% coverage at the moment for `service.package.individually` option in validate.js. This PR adds more unit tests for the missing coverage.

## How did you implement it:

Unit tests are written.

## How can we verify it:

Clone branch, run the test.

## Todos:

- [x] Write tests
- [x] Write documentation
- [x] Fix linting errors

### DIFF
--- a/tests/validate.test.js
+++ b/tests/validate.test.js
@@ -511,7 +511,7 @@ describe('validate', () => {
           globSyncStub.callsFake(filename => [_.replace(filename, '*', 'js')]);
           return expect(module.validate()).to.be.fulfilled
           .then(() => {
-            expect(module.webpackConfig.length).to.equal(4);
+            expect(module.webpackConfig).to.have.lengthOf(4);
             expect(module.webpackConfig[0].output.path).to.equal('output/func1');
             expect(module.webpackConfig[1].output.path).to.equal('output/func2');
             expect(module.webpackConfig[2].output.path).to.equal('output/func3');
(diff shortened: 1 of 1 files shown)


tests/validate.test.js
```diff
@@ -434,6 +434,123 @@ describe('validate', () => {
...
+          });
+        });
+
+        it('should set webpackConfig output path for every functions', () => {
+          module.serverless.service.custom.webpack = testConfig;
+          module.serverless.service.functions = testFunctionsConfig;
+          globSyncStub.callsFake(filename => [_.replace(filename, '*', 'js')]);
+          return expect(module.validate()).to.be.fulfilled
+          .then(() => {
+            expect(module.webpackConfig.length).to.equal(4);
```
HyperBrain: Could be expressed as `expect(module.webpackConfig).to.have.lengthOf(4);` which will implicitly check for an array like structure.

HyperBrain:
@ceilfors Just recognized that this breaks the unit tests on Windows (because of the fixed `/` in the expected paths). I will commit a fix into master to make them work again and use `path.join` to compose the paths.

HyperBrain:
Committed and fixed

ceilfors:
@HyperBrain Apologies! Maybe it's a good idea to add Windows to travis?

HyperBrain:
@ceilfors Yes, I planned to add AppVeyor as second build machine for the unit tests

PR #278, opened 2017-11-15
fix compile stats in multiple packaging

dangcuuson:
## What did you implement:

Fix a problem of [#261](https://github.com/serverless-heaven/serverless-webpack/pull/261#issuecomment-344437763)

## How did you implement it:
Check `multiCompile` flag to extract the correct stats option

## Todos:

- [ ] Write tests

***Is this ready for review?:*** NO: Need to run some manual test/write tests

### DIFF
--- a/lib/compile.js
+++ b/lib/compile.js
@@ -14,18 +14,26 @@ module.exports = {
       .fromCallback(cb => compiler.run(cb))
       .then(stats => {
 
-        if (!this.multiCompile) {
-          stats = { stats: [stats] };
-        }
-
-        const compileOutputPaths = [];
-        const consoleStats = this.webpackConfig.stats || {
+        let consoleStats = {
           colors: true,
           hash: false,
           version: false,
           chunks: false,
           children: false
         };
+
+        if (!this.multiCompile) {
+          stats = { stats: [stats] };
+          if (_.has(this.webpackConfig, 'stats')) {
+            consoleStats = this.webpackConfig.stats;
+          }
+        } else {
+          if (_.has(this.webpackConfig, '0.stats')) {
+            consoleStats = this.webpackConfig[0].stats;
+          }
+        }
+
+        const compileOutputPaths = [];
 
         _.forEach(stats.stats, compileStats => {
           this.serverless.cli.consoleLog(compileStats.toString(consoleStats));


lib/compile.js
```diff
@@ -14,19 +14,27 @@ module.exports = {
       .fromCallback(cb => compiler.run(cb))
       .then(stats => {
 
-        if (!this.multiCompile) {
-          stats = { stats: [stats] };
-        }
-
-        const compileOutputPaths = [];
-        const consoleStats = this.webpackConfig.stats || {
```
HyperBrain: Wouldn't just
```
const consoleStats = this.webpackConfig.stats || _.get(this, 'webpackConfig[0].stats') || { ...
```
be enough instead of the ifs below? Then it would try plain, then array and finally fall back to the defaults.

HyperBrain:
With the one liner above the unit test changes will be minimal too (just one test case). With the big if solution, you have several new code branches that you need to test (even the stats that you initialize now in there).

tests/compile.test.js
```diff
@@ -99,4 +99,31 @@ describe('compile', () => {
       return null;
     });
   });
+
+  it('should use correct stats option', () => {
+    const testWebpackConfig = {
+      stats: 'minimal'
+    };
+    let mockStats = {
```
HyperBrain: This can be const

tests/compile.test.js
```diff
@@ -99,4 +99,31 @@ describe('compile', () => {
...
+    module.webpackConfig = testWebpackConfig;
+    webpackMock.compilerMock.run.reset();
+    webpackMock.compilerMock.run.yields(null, mockStats);
+    return (expect(module.compile()).to.be.fulfilled)
+    .then(() => {
+      module.webpackConfig = [testWebpackConfig];
+      return (expect(module.compile()).to.be.fulfilled);
+    })
+    .then(() => {
+      expect(mockStats.toString.args).to.eql([[testWebpackConfig.stats], [testWebpackConfig.stats]]);
```
HyperBrain: You can do the same for "mockStats.toString.firstCall.args" to check that the plain webpack config's stats was used there.

tests/compile.test.js
dangcuuson: fixed!

tests/compile.test.js
dangcuuson: It already covered in that checks since sinon.args will list the history of all args passed to the spy function, but I have added the first call check. Not sure if excessive test is good or bad :smiling_imp:

dangcuuson:
That's good! :+1:

HyperBrain:
Released with `4.1.0`

tommedema:
@HyperBrain small remaining issue is that this is causing a lot of whitespace:

```






Serverless: Package lock found - Using locked versions
```

With `webpack.config.js` having `stats: 'errors-only'` defined.

HyperBrain:
@tommedema Thanks for the hint 💯 . It might be that 'errors-only' emits an empty console log when printing the stats. Maybe that can be checked for this setting and the log can be skipped. Can you open an issue for that, so that it is tracked?

tommedema:
Thanks, I will create an issue in the near future.

PR #293, opened 2017-12-05
enable webpackIncludeModules for webpack-no-watch

josephleehunsaker:
I got this for webpack-no-watch, not sure how to do this for the watch mode.

### DIFF
--- a/lib/wpwatch.js
+++ b/lib/wpwatch.js
@@ -9,7 +9,9 @@ module.exports = {
     if (this.options['webpack-no-watch']) {
       // If we do not watch we will just run an ordinary compile
       this.serverless.cli.log('Watch disabled by option.');
-      return this.serverless.pluginManager.spawn('webpack:compile');
+      return BbPromise.bind(this)
+        .then(() => this.serverless.pluginManager.spawn('webpack:compile'))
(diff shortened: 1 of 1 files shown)


HyperBrain:
Hi @josephleehunsaker, first thanks for you engagement.

Can you please tell, what exactly the problem is, that this PR should solve? As background: The wpwatch function is only used by serverless-offline which uses the locally available node_modules and does not need an extra packaging step. So the PR will add quite some additional time to the startup (and recompiles if it is added to watch too), eventually rendering local debugging very hard.

If you want to test the final package (as it would be deployed to AWS), you should use `serverless run`. The plugin fully integrates with that, and there the final package will be automatically uploaded into a Serverless Docker image that emulates the API and the lambdas.

josephleehunsaker:
Thx @HyperBrain for the background. So I just use `serverless run` instead of `serverless offline`? Or do I need to run both in separate terminals?

HyperBrain:
It's an either or. If you mant to test the final deployment, use serverless run, if you want to quickly test during development, use serverless offline. The two methods are unrelated and can be used independently.
Asking your question: Just use `serverless run`.

PR #313, opened 2018-01-25
add hook *before:step-functions-offline:start*

vkkis93:
## What did you implement:

Integration **serverless-step-functions-offline** with **serverless-webpack**

## How did you implement it:

Just add hook **'before:step-functions-offline:start'**

## How can we verify it:
Run **serverless-step-functions-offline** plugin in serverless project where **serverless-webpack** will be first runned

```yaml
plugins:
   - serverless-webpack
   ...
   - serverless-step-functions-offline
```

## Todos:

- [ ] Write tests
- [ ] Write documentation
- [x] Fix linting errors

### DIFF
--- a/index.js
+++ b/index.js
@@ -156,7 +156,6 @@ class ServerlessWebpack {
         .then(this.wpwatch),
 
       'before:step-functions-offline:start': () => BbPromise.bind(this)
-         .then(this.validate)
          .then(this.prepareStepOfflineInvoke)
          .then(this.compile)
     };
--- a/index.js
+++ b/index.js
@@ -155,7 +155,7 @@ class ServerlessWebpack {
 
       'before:step-functions-offline:start': () => BbPromise.bind(this)
         .then(this.validate)
-        .then(this.wpwatch)
+        .then(this.compile)
     };
   }
 }
--- a/index.js
+++ b/index.js
@@ -153,6 +153,9 @@ class ServerlessWebpack {
         .then(this.prepareOfflineInvoke)
         .then(this.wpwatch),
 
+      'before:step-functions-offline:start': () => BbPromise.bind(this)
+        .then(this.validate)
+        .then(this.wpwatch)
     };
   }
 }
--- a/index.js
+++ b/index.js
@@ -12,6 +12,7 @@ const run = require('./lib/run');
 const prepareLocalInvoke = require('./lib/prepareLocalInvoke');
 const runPluginSupport = require('./lib/runPluginSupport');
 const prepareOfflineInvoke = require('./lib/prepareOfflineInvoke');
+const prepareStepOfflineInvoke = require('./lib/prepareStepOfflineInvoke');
 const packExternalModules = require('./lib/packExternalModules');
 const packageModules = require('./lib/packageModules');
 const lib = require('./lib');
@@ -44,7 +45,8 @@ class ServerlessWebpack {
       packageModules,
       prepareLocalInvoke,
       runPluginSupport,
-      prepareOfflineInvoke
+      prepareOfflineInvoke,
+      prepareStepOfflineInvoke
     );
 
     this.commands = {
@@ -154,8 +156,9 @@ class ServerlessWebpack {
         .then(this.wpwatch),
 
       'before:step-functions-offline:start': () => BbPromise.bind(this)
-        .then(this.validate)
-        .then(this.compile)
+         .then(this.validate)
+         .then(this.prepareStepOfflineInvoke)
+         .then(this.compile)
     };
   }
 }
--- a/lib/prepareStepOfflineInvoke.js
+++ b/lib/prepareStepOfflineInvoke.js
@@ -8,17 +8,17 @@ const path = require('path');
  */
 
 module.exports = {
-    prepareStepOfflineInvoke() {
-        _.set(this.serverless, 'service.package.individually', false);
+  prepareStepOfflineInvoke() {
+    _.set(this.serverless, 'service.package.individually', false);
 
-        return this.serverless.pluginManager.spawn('webpack:validate')
-            .then(() => {
-                if (!this.options.location && !_.get(this.serverless, 'service.custom.stepFunctionsOffline.location')) {
-                    _.set(this.serverless, 'service.custom.stepFunctionsOffline.location',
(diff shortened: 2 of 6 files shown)


HyperBrain:
Hey @vkkis93 , nice addition 👍 Will be part of the next release.

Are there any caveats if using step-functions offline together with serverless-offline, or is that not possible?

index.js
```diff
@@ -153,6 +153,9 @@ class ServerlessWebpack {
         .then(this.prepareOfflineInvoke)
         .then(this.wpwatch),
 
+      'before:step-functions-offline:start': () => BbPromise.bind(this)
+        .then(this.validate)
+        .then(this.compile)
```
HyperBrain: Watch will automatically watch for source changes. Both functionalities (offline watch and invoke local watch) set the base directory for the execution properly, e.g. for serverless-offline the location parameter is set correctly.

If there is something similar for the step-functions that needs to be done, we had to add a different "watch" function for step (e.g. stepWatch).

vkkis93: Instead of  using .then(this.validate).then(this.compile) need to create separate function for plugin ?

HyperBrain: Only _if_ we want to support watch mode and _if_ step-offline needs to know the compiled directory, then you can do:
```
.then(this.validate)
.then(this.prepareStepOffline)   <<<--- see here
.then(this.wpwatch)
```

The new `prepareStepOffline` function should then do something similar as https://github.com/serverless-heaven/serverless-webpack/blob/master/lib/prepareOfflineInvoke.js, only that it prepares the root folder for the step-offline plugin.

vkkis93: Need only compile code once, run step-offline and stop process.
I don't need to have watch mode
About compiled directory, user will indicate the path to directory by himself.

Referring to my requirements - need to change something or this code makes sense?

HyperBrain: Makes sense with the compile, yes.

However:

> About compiled directory, user will indicate the path to directory by himself.

That will not really work out, if you set package: individually. Then the compiled code will then be distributed. I think you need a "prepare" function because for the offline, it has to be set to service packaging (then the output directory is ".webpack/service") like in the prepareOffline.

With individual packaging it is ".webpack/<function-name>" which is not predictable.

vkkis93: Sorry, I mean compiled files.
In my plugin user specifies path to function directly
```yaml
custom:
  stepFunctionsOffline:
    FirstLambda: firstLambda/index.handler
```

in case with compiled files it will look like

```yaml
custom:
  stepFunctionsOffline:
    FirstLambda: firstLambda/index.handler
    SecondLambda: .webpack/src/index.handler // in case  individually: true
    SecondLambda: .webpack/service/src/index.handler
```
step-offline does not work on another.

HyperBrain: `SecondLambda: .webpack/src/index.handler // in case  individually: true`

That is wrong, because the function name is contained in the path. The service packaging path is ok, so it would be better if the webpack plugin would automatically enable service packaging for offline mode - exactly as it is done with serverless-offline.
Then it will work correctly all the time.

Additionally, step-offline could have a location parameter (same as serverless-offline). Then it would be cool and the user does not have to do any own configuration.

vkkis93: I mean .webpack/{functionName}/src/index.handler

with your approach I need to rewrite my plugin quite a lot

HyperBrain: Can you give me the link to your plugin. Maybe I have some time over the weekend to have a look, if we can combine this in an easy and compatible way there.

vkkis93: @HyperBrain but looks like your approach very good.
Makes sense to rewrite it.
Only one question, serverless-webpack compiles all functions which described in section *functions* ?

HyperBrain: Yes

vkkis93: Great, 
I update  this pull request to *Is this ready for review?: NO* at this moment.
When I will rewrite code, I will notify you

HyperBrain: Good idea. I'm already excited 😄

vkkis93:
Hi @HyperBrain .
I have done fix and now PR is ready for review.
The only one problem I see it's coveralls.
Need to write some tests on my new file ?

index.js
```diff
@@ -153,6 +155,10 @@ class ServerlessWebpack {
         .then(this.prepareOfflineInvoke)
         .then(this.wpwatch),
 
+      'before:step-functions-offline:start': () => BbPromise.bind(this)
+         .then(this.validate)
```
HyperBrain: You already spawn the `webpack:validate` in the prepareStepOfflineInvoke function, so you can remove the line here.

vkkis93:
@HyperBrain  can you show the example of unit tests for serverless-offline ?

index.js
vkkis93: ahh, you're right.
I missed that

HyperBrain:
Oops. Just saw that these unit test are also missing.

I will add the prepare unit tests for StepOffline to this PR, so you don't have to take care of them 😄

vkkis93:
@HyperBrain great news :relaxed:

vkkis93:
Hi @HyperBrain .
What should be done in order to close this PR?

HyperBrain:
I'll finish it tomorrow and merge it. Sorry for the delay, was busy over the weekend ;-)

lib/prepareStepOfflineInvoke.js
```diff
@@ -0,0 +1,24 @@
+'use strict';
+
+const _ = require('lodash');
+const path = require('path');
+
+/**
+ * Special settings for use with serverless-offline.
```
j0k3r: This should be `step-functions-offline` instead of `serverless-offline`.

vkkis93: Fixed

vkkis93:
Thanks a lot

HyperBrain:
The new implementation is now covered 100%. The decrease is because a different unrelated file is not correctly tested. I will merge this PR now and fix the other unit test in a separate PR.
@vkkis93 Thanks again for the PR 👍

j0k3r:
@vkkis93 Good job!

jazen:
:+1: Looking forward to `4.4.0`

HyperBrain:
Released with `4.4.0`

PR #318, opened 2018-02-04
Add warning about auto-packing when using yarn

### LINKED ISSUE #316: Fix/discourage use of Yarn with serverless-webpack
# This is a Bug Report

## Description

We have a fairly straightforward serverless app that uses this package, `yarn` for module installation and the externals auto-packing config as described here https://github.com/serverless-heaven/serverless-webpack#node-modules--externals . 

We deploy our app from CircleCi which uses node 6.10.3/npm 3.x. We have a lambda that has npm dependencies, and within `node_modules` they are being hoisted as they are shared. When `serverless-webpack` does auto-packing it runs an `npm install` which results in an incomplete package - the lambda fails because the hoisted dependencies are not there. 

Our fix was to upgrade to npm 5.x on CircleCI, which creates the `package-lock.json`, but really it would be great if the docs were explicit that this is going on behind the scenes - or ideally, support yarn.

Happy to help with this and many thanks for your hard work.

## Additional Data

* ***Serverless-Webpack Version you're using***: 4.0.0
* ***Webpack version you're using***: 3.8.1
* ***Serverless Framework Version you're using***: 1.24.0
* ***Operating System***: Ubuntu 14.04

dpiatek:
## What did you implement:

Closes #316 

## How did you implement it:

Docs update.

## How can we verify it:

Check statements and spelling 👨‍🏫 

## Todos:

- [x] Write documentation

### DIFF
--- a/README.md
+++ b/README.md
@@ -210,6 +210,12 @@ custom:
 ```
 > Note that only relative path is supported at the moment.
 
+#### Usage with yarn
+
+Note that if auto-packing is enabled, the plugin will call `npm install`. If you are using yarn your `yarn.lock` file will be not be honored, which might lead to unexpected results as your dependencies will most likely not match (or be missing, as npm does not install packages in the same way as yarn).
+
+Yarn support is planned [#286][link-286]. Until then we recommend using npm when using auto-packing.
+
 #### Forced inclusion
 
 Sometimes it might happen that you use dynamic requires in your code, i.e. you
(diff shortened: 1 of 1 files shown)


HyperBrain:
@dpiatek 👍 That should be good to let people know about the issue upfront.

HyperBrain:
BTW: `Check statements and spelling 👨‍🏫` ... I'm the wrong one to check the spelling (non-native speaker) but I saw the new statements 😄

README.md
```diff
@@ -210,6 +210,12 @@ custom:
 ```
 > Note that only relative path is supported at the moment.
 
+#### Usage with yarn
+
+Note that if auto-packing is enabled, the plugin will call `npm install`. If you are using yarn your `yarn.lock` file will be not be honored, which might lead to unexpected results as your dependencies will most likely not match (or be missing, as npm does not install packages in the same as npm).
```
HyperBrain: Shouldn't this be `[...] install packages in the same *way as yarn*`?

dpiatek: Can't believe I did not see that, fixed!

Issue #315, opened 2018-01-31
Question: hooks *after* compile – but *before* run


dwelch2344:
# This is a Question

## Description

New to plugin development, so apologies if we're missing something incredibly obvious. We're attempting to write a plugin that runs after webpack but before deploy / offline / etc, but can't seem to hit the hook. We can run after `validate`, just fine (with `after:webpack:validate:validate`)

Plugin thus far is dead simple. Uncomment the validate hook to see it in action. So far the plugin is just running in `.serverless_plugins` – and is imported *after* `serverless-webpack` (though we've tried before as well)

Are we missing something obvious? Or is there not a hook for the compilation?

Thanks

```js
'use strict';

class ServerlessPlugin {
  constructor(serverless, options) {    
    this.serverless = serverless;
    this.options = options;

    this.commands = {
      welcome: {
        usage: 'Helps you start your first Serverless plugin',
        lifecycleEvents: [
          'hello',
        ],  
        commands: {
          compile: {
            type: 'entrypoint',
            lifecycleEvents: [
              'compile',
            ],
          },
        }
      },      
    };

    this.hooks = {
      // 'after:webpack:validate:validate': this.validate.bind(this),
      'webpack:compile:compile': this.debug.bind(this),
      // 'before:offline:start': this.debug.bind(this),
      // 'offline:start': this.debug.bind(this),
      'before:offline:start': this.debug.bind(this),
      // 'before:offline:start:init': this.debug.bind(this),      
    };
  }

  validate(){
    for(let i = 0; i < 15; i++){
      console.log('validate')
    }
  }

  debug(){
    throw new Error('i should blow up here debug')
  }


  compile(){
    throw new Error('i should blow up here compile')
  }
}

module.exports = ServerlessPlugin;

```

## Additional Data

* ***Serverless-Webpack Version you're using***:
^4.2.0
* ***Webpack version you're using***:
^3.10.0
* ***Serverless Framework Version you're using***:
^1.25.0
* ***Operating System***:
Mac OS X - Sierra (scared to upgrade to High Sierra)

HyperBrain:
Hi @dwelch2344 , thanks for asking.

This hook is wrong:
`'webpack:compile:compile': this.debug.bind(this),`
because it "overwrites" the compile step (in fact the bahavior is undefined in this case because an overwrite/replacement is currently not supported by Serverless). 

In case you want to hook into some serverless lifecycle event, you always have to hook besides the offered hook. In your case this would be:
[...]

PR #319, opened 2018-02-05
#315 - Add watch-compile event

dwelch2344:
Quick shoutout: I'm new to plugin development and not sure the best way to test / lint / etc, but happy to make changes and do the legwork as needed.

## What did you implement:

We've got a plugin in development that needs access to the post-compiled code for our handler, but in local development but there's no way to detect compiled changes if using watch. We've toyed with adding a new event on `lib/wpwatch`
that plugins could hook on as needed. 

See an example project at https://github.com/dwelch2344/serverless-webpack-watch-compile-demo and the issue #315 
filed at https://github.com/serverless-heaven/serverless-webpack/issues/315

## How did you implement it:

Added a `watch-compile` definition in `index.js` and a `serverless.spawn` of said event in `lib/wpwatch.js` 

## How can we verify it:

* Install locally to your npm 
* Clone this project: https://github.com/dwelch2344/serverless-webpack-watch-compile-demo
* Adjust the project's `package.json` to use your local version
* run `npm run debug` and observe the plugin logs 15 times:
```
→ npm run debug

> serverless-webpack-watch-compile@1.1.0 debug /tmp/serverless-webpack-watch-compile
> node --inspect ./node_modules/.bin/serverless offline -P 5000 -s dev

Debugger listening on ws://127.0.0.1:9229/8ee583a5-086a-407b-829d-c1a971a4d39f
For help see https://nodejs.org/en/docs/inspector
Serverless: Running after a compiled change detected
Serverless: Running after a compiled change detected
Serverless: Running after a compiled change detected
Serverless: Running after a compiled change detected
Serverless: Running after a compiled change detected
Serverless: Running after a compiled change detected
Serverless: Running after a compiled change detected
Serverless: Running after a compiled change detected
Serverless: Running after a compiled change detected
Serverless: Running after a compiled change detected
Serverless: Running after a compiled change detected
Serverless: Running after a compiled change detected
Serverless: Running after a compiled change detected
Serverless: Running after a compiled change detected
Serverless: Running after a compiled change detected
Time: 519ms
         Asset     Size  Chunks             Chunk Names
    handler.js  5.52 kB       0  [emitted]  handler
handler.js.map  5.38 kB       0  [emitted]  handler
   [0] ./handler.js 2.35 kB {0} [built]
   [1] external "babel-runtime/core-js/promise" 42 bytes {0} [not cacheable]
   [2] external "babel-runtime/helpers/objectWithoutProperties" 42 bytes {0} [not cacheable]
   [3] external "babel-runtime/regenerator" 42 bytes {0} [not cacheable]
   [4] external "babel-runtime/core-js/json/stringify" 42 bytes {0} [not cacheable]
   [5] external "babel-runtime/helpers/asyncToGenerator" 42 bytes {0} [not cacheable]
   [6] external "source-map-support/register" 42 bytes {0} [not cacheable]
Serverless: Watching for changes...
Serverless: Starting Offline: dev/us-east-1.

Serverless: Routes for hello:
Serverless: GET /hello

Serverless: Offline listening on http://localhost:5000
```

* Now make a change in `handler.js`, and observe the plugin logs again 

```
Time: 88ms
         Asset     Size  Chunks             Chunk Names
    handler.js  5.52 kB       0  [emitted]  handler
handler.js.map  5.38 kB       0  [emitted]  handler
   [0] ./handler.js 2.36 kB {0} [built]
   [1] external "babel-runtime/core-js/promise" 42 bytes {0} [not cacheable]
   [2] external "babel-runtime/helpers/objectWithoutProperties" 42 bytes {0} [not cacheable]
   [3] external "babel-runtime/regenerator" 42 bytes {0} [not cacheable]
   [4] external "babel-runtime/core-js/json/stringify" 42 bytes {0} [not cacheable]
   [5] external "babel-runtime/helpers/asyncToGenerator" 42 bytes {0} [not cacheable]
   [6] external "source-map-support/register" 42 bytes {0} [not cacheable]
Serverless: Watching for changes...
Serverless: Running after a compiled change detected
Serverless: Running after a compiled change detected
Serverless: Running after a compiled change detected
Serverless: Running after a compiled change detected
Serverless: Running after a compiled change detected
Serverless: Running after a compiled change detected
Serverless: Running after a compiled change detected
Serverless: Running after a compiled change detected
Serverless: Running after a compiled change detected
Serverless: Running after a compiled change detected
Serverless: Running after a compiled change detected
Serverless: Running after a compiled change detected
Serverless: Running after a compiled change detected
Serverless: Running after a compiled change detected
Serverless: Running after a compiled change detected
```

* Verify your change is there and things worked as expected

## Todos:

- [ ] Write tests
- [ ] Write documentation
- [ ] Fix linting errors

### DIFF
--- a/index.js
+++ b/index.js
@@ -71,12 +71,14 @@ class ServerlessWebpack {
             lifecycleEvents: [
               'compile',
             ],
-          },
-          "watch-compile": {
-            type: 'entrypoint',
-            lifecycleEvents: [
-              'watch-compile',
-            ],
+            commands: {
+              watch: {
+                type: 'entrypoint',
+                lifecycleEvents: [
+                  'compile'
+                ]
+              }
+            }
           },
           package: {
             type: 'entrypoint',
@@ -144,6 +146,8 @@ class ServerlessWebpack {
 
       'webpack:compile:compile': () => BbPromise.bind(this)
         .then(this.compile),
+
+      'webpack:compile:watch:compile': () => BbPromise.resolve(),
 
       'webpack:package:packExternalModules': () => BbPromise.bind(this)
         .then(this.packExternalModules),
--- a/index.js
+++ b/index.js
@@ -72,6 +72,12 @@ class ServerlessWebpack {
               'compile',
             ],
           },
+          "watch-compile": {
+            type: 'entrypoint',
+            lifecycleEvents: [
+              'watch-compile',
+            ],
+          },
           package: {
             type: 'entrypoint',
             lifecycleEvents: [
--- a/lib/wpwatch.js
+++ b/lib/wpwatch.js
@@ -45,7 +45,8 @@ module.exports = {
 
         process.env.SLS_DEBUG && this.serverless.cli.log(`Webpack watch invoke: HASH NEW=${stats.hash} CUR=${lastHash}`);
 
-        // If the file hash did not change there were no effective code changes detected (comment only changed do not change the compile hash!)
+        // If the file hash did not change there were no effective code changes detected
+        // (comment changes do not change the compile hash and do not account for a rebuild!)
         // See here: https://webpack.js.org/api/node/#watching (note below watching)
         if (stats && stats.hash === lastHash) {
           if (firstRun) {
--- a/lib/wpwatch.js
+++ b/lib/wpwatch.js
@@ -52,7 +52,7 @@ module.exports = {
           firstRun = false;
           callback();
         }else{
-          this.serverless.pluginManager.spawn('webpack:compile:watch')
+          this.serverless.pluginManager.spawn('webpack:compile:watch');
         }
       });
     };
--- a/lib/wpwatch.js
+++ b/lib/wpwatch.js
@@ -31,9 +31,10 @@ module.exports = {
     };
 
     // This starts the watch and waits for the immediate compile that follows to end or fail.
+    let lastHash = null;
     const startWatch = (callback) => {
       let firstRun = true;
-      compiler.watch(watchOptions, (err, stats) => {
+      const watcher = compiler.watch(watchOptions, (err, stats) => {
         if (err) {
           if (firstRun) {
             firstRun = false;
@@ -42,17 +43,39 @@ module.exports = {
           throw err;
         }
 
+        process.env.SLS_DEBUG && this.serverless.cli.log(`Webpack watch invoke: HASH NEW=${stats.hash} CUR=${lastHash}`);
+
+        // If the file hash did not change there were no effective code changes detected (comment only changed do not change the compile hash!)
+        // See here: https://webpack.js.org/api/node/#watching (note below watching)
+        if (stats && stats.hash === lastHash) {
+          if (firstRun) {
+            firstRun = false;
+            callback();
+          }
+          return;
+        }
+
         if (stats) {
+          lastHash = stats.hash;
           this.serverless.cli.consoleLog(stats.toString(consoleStats));
         }
 
-        this.serverless.cli.log('Watching for changes...');
-
         if (firstRun) {
           firstRun = false;
+          this.serverless.cli.log('Watching for changes...');
           callback();
-        }else{
-          this.serverless.pluginManager.spawn('webpack:compile:watch');
+        } else {
+          // We will close the watcher while the compile event is triggered and resume afterwards to prevent race conditions.
+          watcher.close(() => {
+            return this.serverless.pluginManager.spawn('webpack:compile:watch')
+            .then(() => {
+              // Resume watching after we triggered the compile:watch event
+              return BbPromise.fromCallback(cb => {
+                startWatch(cb);
+              })
+              .then(() => this.serverless.cli.log('Watching for changes...'));
+            });
+          });
         }
       });
     };
--- a/lib/wpwatch.js
+++ b/lib/wpwatch.js
@@ -52,7 +52,7 @@ module.exports = {
           firstRun = false;
           callback();
         }else{
-          this.serverless.pluginManager.spawn('webpack:watch-compile')
+          this.serverless.pluginManager.spawn('webpack:compile:watch')
         }
       });
     };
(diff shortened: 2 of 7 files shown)


index.js
```diff
@@ -72,6 +72,12 @@ class ServerlessWebpack {
               'compile',
             ],
           },
+          "watch-compile": {
```
HyperBrain: You should move that as subcommand to the compile command (as entrypoint):
```js
compile:
  ...
  commands: {
    watch: {
      type: 'entrypoint',
      lifecycleEvents: [
        'compile'
      ]
    }
  }
```
Then it is more universal and better located semantically.

Additionally add a hook to define the new hookable event in `this.hooks` like this:
```js
this.hooks = [
  ...
  'webpack:compile:watch:compile': () => BbPromise.resolve();
};
```
so that it is owned by the plugin and can be addressed with
`after:webpack:compile:watch:compile` and the symmetric before.

lib/wpwatch.js
```diff
@@ -51,6 +51,8 @@ module.exports = {
         if (firstRun) {
           firstRun = false;
           callback();
+        }else{
+          this.serverless.pluginManager.spawn('webpack:watch-compile')
```
HyperBrain: This will be `webpack:compile:watch` then

HyperBrain:
If you need any help, please let me know.

dwelch2344:
Great feedback, thank you. Only one question: 
> And you should spawn also for the other watch commands (with invoke local and run).
Not sure I follow exactly. What does that entail / look like? 

Again, fairly new to plugin development (and Serverless as a whole, actually) – but happy to do the legwork + reading if you can point me in the right direction / give me a few examples. 

Thanks!

HyperBrain:
The other kinds of watches are (quite similar code) in `run.js`

HyperBrain:
... and no problem that you're a newcomer to plugin development 😃 . I'm here to guide you - and if you have generic questions about the Serverless plugin system, it's also no problem for me to help.

dwelch2344:
@HyperBrain  Okay, I think we're good. I've moved to the subcommand (nifty – definitely like those) and verified things still work as expected. Also adjusted tests in both `run.test.js` as well as `wpwatch.test.js` and coverage has actually gone up :) 

Regarding spawning on the other watch commands: I think this particular hook is right where it needs to be, as it really is only on subsequent compiles via watch that we want to hook in the event (and thus the tests ensuring we don't spawn with 1 invocation but do with 2+ invocations). Is there a gap I'm missing / flaw in my thinking there?

HyperBrain:
@dwelch2344 Cool, I'll have a look later 😄 

The current implementation you did is fine and should serve its purpose.

The only thing I mentioned before was this (maybe minor and optional thing):
With awaiting the spawn I meant that now the hooks are triggered by the watch implementation - which is perfectly done how it is. But the spawn() itself is not awaited within the watch callback, what means that it can happen that the hook is triggered, dependent plugins do something, but as the callback already returned, an additional hook invocation can happen while the previous one is not yet finished.

That might actually never happen ;-) - or can be fixed _if_ it happens.

HyperBrain:
@dwelch2344 Just saw that the new hookable event should be added to the README (section "for developers"). Can you do that?

HyperBrain:
I just restarted the build (there is something not ok with the `validate` unit tests - but that's out of scope here). Now it succeeded.

dwelch2344:
@HyperBrain Ah thank you for the clarification! Yeah, that definitely makes sense and we should do that to keep things running in the right / clean order. Would hate to leave a landmine for someone (especially since we'll be using this soon 😄 )

So, to await for it, do I simply need to "return" the spawn invocation? Assuming it's a promise, and since it's the last call in this instance,  I'd think that would do it (based on the other spawn calls above). Is that accurate? If not, mind pointing out an example?

HyperBrain:
> do I simply need to "return" the spawn invocation?

I think that will not work, because the function we're calling the spawn in is webpack's watch callback - and I doubt that webpack will await it. We might need some more sophisticated stuff here, that will only return from the callback when the spawn returns.

Something that is similar a lock/wait mechanism in other languages, that blocks the callback function until it is released. I'll try to investigate a proper solution tomorrow that handles a wait for a promise in a synchronous function. In theory the callback function itself must be wrapped somehow .... but let's see

dwelch2344:
Ok. Quick question then: how do releases on this project work? Aka how long would it take to get this out into a release?  If they're small / fast cycles for this kind of thing, could we cut a release and open an enhancement to do the proper solution (which I'm happy to continue work on)? 

We've got a pretty nifty setup and this is the last piece we need to continue rolling our product out, so while we're monkey-patching it now to dev it'd be nice to sew it up asap. Which I realize is trying to make our problem yours, but figured worth asking :smile:

Thanks

HyperBrain:
@dwelch2344 I commited a fix for the race condition and tested it locally with `serverless offline`. Now the event code is awaited correctly and watching resumes after the events have been processed.
Additionally I fixed a generic issue with unnecessary recompiles by checking the webpack compile hash as proposed on their site.
Can you pull the branch and do some tests too? We have to make sure that it now works expected before it can be merged.

I have to adapt the unit tests now and change the run watcher in the same way. Afterwards I'll merge if your tests were successful.

Regarding the release schedule: There are no fixed release cycles and I generally release, as soon as there are important things available and ready for the next minor or patch release.

dwelch2344:
This is awesome! Thanks for being so responsive 👍  Will check out and validate ASAP, EOD at the latest. 

On another note: would love to help contribute and learn from your plugin experience. Could we connect via another channel? Gtalk / twitter / slack / gitter? I'll come to you :)

dwelch2344:
Just confirmed! this looks great! Can we go with it?

HyperBrain:
Yes we can 😄 . I'll just check the local invoke watch and the run watch, then we're good to go.

... and of course we can connect - I assume you're in PST, so it's 9 hours difference but still enough overlap :-D

dwelch2344:
Actually I'm in MT, so 10 probably ;)  But I keep odd hours often so happy to work around you. DM me your preferred contact method on twitter at twitter.com/david_welch ?

HyperBrain:
I'll postpone the event trigger for local invoke and run to a second PR, because that implementation has to be cleaned up anyway.
You'll hear from me via DM this weekend ;-)

PR #327, opened 2018-02-26
Support package.include and package.exclude

### LINKED ISSUE #394: Exclude Readme and tests file in external modules
I do not want external modules to be bundled, but want them to be packaged in artifacts. I can do this by custom option 
```custom:
  webpack:
    includeModules: true```
 But this plugin includes all files in the external module. How do I exclude Read me and test files from external modules in the packaged artifact?

### PR DESCRIPTION
**Pull request by janicduplessis**
## What did you implement:

Closes #394 

This makes sure to respect the [serverless package.include and package.exclude](https://serverless.com/framework/docs/providers/aws/guide/packaging#exclude--include) options when bundling.

## How did you implement it:

Simple port of the packaging code that handles these option in serverless (https://github.com/serverless/serverless/blob/ba979a7a8fc6b095910a2cd67cb6413ac7524952/lib/plugins/package/lib/packageService.js#L130).

Also changed the glob dependency for globby since it's what serverless is using and made the port easier (supports an array of glob). It was only used at one other place in the project.

## How can we verify it:

- Tested the change in a project by ignoring source maps generated by webpack (package.exclude: "*.map") and verifying the resulting zip did not contain the files. Also tested that it still works like before when not setting any of these options.
- Added a test that asserts the correct glob pattern is passed.

## Todos:

- [x] Write tests
- [ ] Write documentation
- [x] Fix linting errors

***Is it a breaking change?:*** NO, debatable since the option was ignored before but I don't think so.

### CODE REVIEW

File: lib/validate.js [hunk 1]
```diff
@@ -30,8 +30,8 @@ module.exports = {
     };
 
     const getEntryExtension = fileName => {
-      const files = glob.sync(`${fileName}.*`, {
-        cwd: this.serverless.config.servicePath,
+      const files = globby.sync(`${fileName}.*`, {
+        cwd: this.serverless.config.servicePath || process.cwd(),
```
HyperBrain: I'm not sure if defaulting to CWD is right here. It will miss the handlers in case CWD !== servicePath and render the webpack configuration invalid. In general, serverless-webpack depends on servicePath everywhere and uses it to create it's outputs.

HyperBrain:
Hi @janicduplessis . The packager works on the compiled output, which is built and defined by webpack. Excluding or including stuff there might break integrity. Including special files can be easily done the webpack-way by using the CopyFiles plugin.

Thanks for the PR. Can you elaborate on the concrete use case that is solved by enabling the package includes and excludes? Maybe I just miss the use-case here 😄

lib/validate.js
janicduplessis: I added this since `servicePath` was null in some cases (maybe just in tests) and globby doesn't support that while glob did. I verified that both libs default to process.cwd if cwd is not set so this should be fine.

HyperBrain: Yes, that can be the tests. Not all of them intialize a full config object for testing. So this should be ok 👍

janicduplessis:
@HyperBrain Hello! The concrete use case for me is excluding source maps from the generated bundle. I still want webpack to generate them so I can upload them to crash reporting services but I don't want to add to the function size.

I assumed this option could solve this problem cleanly while still remaining somewhat generic. I agree we operate on compiled output there so it is a bit different than how it's use normally but it seems to work well to ignore specific file types. AFAIK the option isn't and couldn't be used for something else so I feel like it makes sense to reuse it here.

lib/validate.js
HyperBrain: BTW: I just remember another open PR, that stabilizes the directories and uses webpackconfig.context to keep the webpack root directory. So this line would be void if the two PR are merged together.
https://github.com/serverless-heaven/serverless-webpack/pull/325

homerjam:
Hi both, I've just come across a reason this PR would be useful. I'm using docker to `npm rebuild` linux bindings and need to include these explicitly using the `package:include:` option in `serverless.yml`. I'll be trying the CopyFiles plugin as an alternative for now.

Thanks

HyperBrain:
@janicduplessis @homerjam I though about the package include and exclude again. Reusing the properties from Serverless might be super confusing, because in the plugin's context they will only operate on the compiled code, but not on the sources. For webpack, the right way to include files from outside of the compile output directories is to use proper webpack plugins, like the file-loader or the copyfiles plugin.

However I think that the settings make sense and should be there for the compiled code.

What, if we put the settings under a new `custom: webpack: package:` property, to make visible that it is something different? I already had the plan to consolidate all webpack specific configuration settings under the `custom: webpack:` hive anyway.

homerjam:
Sounds like a good idea to me. To be honest I think it'd help just to be clearer - I ended up here because I was confused why my includes weren't being included all of a sudden, having added serverless-webpack to an existing project. Now I understand it makes total sense!

homerjam:
As an aside what do you think of including something like [this](https://gist.github.com/jokeyrhyme/d57097a491aa5ecaf27532d057d72461)? As this plugin is installing from npm (right?) it could go one further and build the bindings for linux too (if docker is available)?

HyperBrain:
Info: I'll come back to this one soon after the 5.1.0 release.

HyperBrain:
@homerjam To the aside you mentioned. Now with version 5.x having full support for arbitrary packagers, a Docker based one that would just use Docker to package the external dependencies would be possible. However this should be discussed in a separate issue, to see what others think about that and how exactly it should be implemented to fulfill the needs.

janicduplessis:
@HyperBrain Sorry about taking so long to get back on this, I moved the config to a separate field, let me know if that looks good and all that's left to do is update the doc.

lib/packageModules.js
```diff
@@ -34,11 +34,27 @@ function zip(directory, name) {
 
   const output = fs.createWriteStream(artifactFilePath);
 
-  const files = glob.sync('**', {
+  const exclude = _.get(this.configuration, 'package.exclude', []);
```
HyperBrain: The configuration is now separated into the `Configuration` object. You should add a getter for `package` and set a viable default (empty array) _there_ and add some unit tests. Then just use `const { exclude, include } =  this.configuration.package` here instead of bypassing the integrity of the class.

tests/packageModules.test.js
```diff
@@ -319,6 +319,56 @@ describe('packageModules', () => {
       });
     });
 
+    it('should respect package.include and package.exclude', () => {
+      _.set(module, 'configuration.package', {
```
HyperBrain: Same here accordingly: ` _.set(module, 'configuration._config.package', {`

stevemao:
Would be good if `package.include` works the same way as `CopyFiles` plugin. For people just added this plugin they don't need to add `CopyFiles` plugin and it still works the same way as before.

xyklex:
What is left for this feature to be merged, only the changes @HyperBrain mentioned?

jpoissant:
Any news on this merge?

simon-casasoft:
Any update on this please?

nponeccop:
There are merge conflicts. The PR will be merged once someone resolves them. Anyone can do it. I will do it once I have time.

janicduplessis:
I don't think this is needed anymore, and no longer have a use case for it. If this is still useful to someone feel free to pickup this PR.

PR #328, opened 2018-02-26
Use chunk.forEachModules instead of deprecated chunk.modules

janicduplessis:
## What did you implement:

Webpack 4 removes the deprecated chunk.modules property. This uses the forEachModule instead. See https://github.com/webpack/webpack/blob/webpack-3/lib/Chunk.js#L473.

`forEachModule` exists since webpack v3 so this means webpack v2 is no longer supported with this change. Not sure how much we care about webpack 2 support. If we want we could check if `forEachModule` exists and fallback to `.modules`.

## How did you implement it:

Trivial, most changes were to update tests.

## How can we verify it:

- Tested in a project using webpack 4, before this change externals were broken.
- Run tests

## Todos:

- [x] Write tests (N/A)
- [x] Write documentation (N/A)
- [x] Fix linting errors

***Is it a breaking change?:*** YES (if webpack 2 is still supported)


HyperBrain:
Hi @janicduplessis, thanks for the PR. It indeed makes sense to switch to the no-deprecated chunk module enumerator.

Regarding webpack 2 support, I think there are a lot of serverless projects out there that make use of it. Either we schedule this change for `serverless-webpack@5.0.0` and drop webpack 2 support completely or we implement a fallback mechanism.

However, I'd prefer a new major version, because as the old functionality is deprecated, it is much cleaner to continue with only the correct implementation and drop webpack 2 support.

janicduplessis:
@HyperBrain I also think scheduling this for 5.0.0 is better. It would add support for webpack@4 and drop webpack@2.

HyperBrain:
I retargeted this PR to the new v5 branch. According to #331 this is urgently needed to support webpack v4, so we'll prepare the v5 branch to gather everything needed to support it.

@janicduplessis Did you try already, if this PR enabled webpack 4 support?

HyperBrain:
> Tested in a project using webpack 4, before this change externals were broken.

Sorry 😄 . Did not read that.
The PR looks good to me, so I'll merge it into the v5 branch now.

HyperBrain:
Released with 5.0.0

PR #336, opened 2018-03-07
Refactor configuration

HyperBrain:
## What did you implement:

This PR consolidates all webpack plugin configuration settings into the new `custom.webpack` object. It effectively reduces cluttering of the serverless configuration and keeps all settings in one place.

There is one breaking change with the configuration: If you have chosen to embed the webpack configuration into the serverless.yml (previously by defining `custom.webpack` as an object with the contents of a webpack config file) you have to move that to `custom.webpack.config`. In my opinion this way to configure webpack is very rare and mainly used as convenience method for the unit tests, as it does not allow for dynamic entry point resolution. This feature will be dropped in the future.

The new configuration object has the following layout. If a property is omitted, it will default to the shown values:
```yaml
custom:
  webpack:
    webpackConfig: 'webpack.config.js'
    webpackIncludeModules: false
    packager: 'npm'
    packExternalModulesMaxBuffer: 200 * 1024
}
```
All settings have just been moved, but semantically are exactly as before.

However, **the old configuration keys continue to work**, but will lead to a warning message to let people know that there is a better alternative for configuration now.

## How did you implement it:

Use a new class to keep the configuration centralized. The class is initialized with the serverless custom object and will populate the settings from there (using the old semantics as fallback) and it will set reasonable defaults for anything that has not been set.

## How can we verify it:

(1) Use an existing project as is. It should build as usual but show a warning.
(2) Adapt the configuration to use the new custom.webpack object. It should build as before.

## Todos:

- [x] Write tests
- [x] Write documentation
- [x] Fix linting errors
- Make sure code coverage hasn't dropped **N/A**
   This PR introduced an expected drop in coverage, because it enables counting _all_ files for coverage calculation. Previously only the tested ones were counted, but it was not obvious that there were untested files in the project.

***Is it a breaking change?:*** YES (most likely not noticable)

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,6 +29,7 @@ WebPack's [Tree-Shaking][link-webpack-tree] optimization.
 ## Recent improvements and important changes
 
 * Webpack 2 support has been dropped in favor of Webpack 4
+* Cleaned up configuration. You should now use a `custom.webpack` object to configure everything relevant for the plugin. The old configuration still works but will be removed in the next major release. For details see below.
 * This 5.0.0 prerelease is based on the current 4.4.0
 
 For the complete release notes see the end of this document.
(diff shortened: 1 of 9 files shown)


HyperBrain:
Released with 5.0.0

PR #340, opened 2018-03-08
Yarn support

### LINKED ISSUE #286: Support optional use of Yarn for packaging (instead of NPM)
Hey! 👋 

As discussed [here](https://github.com/trek10inc/serverless-secrets/issues/26#issuecomment-346095775) with @HyperBrain, it'd be great if we could optionally use Yarn for packaging files, instead of NPM. 🙂 

In particular, since `serverless-webpack` completely relies on NPM and uses it for packaging files, I'm having issues using `serverless-webpack` with [Yarn Workspaces](https://yarnpkg.com/blog/2017/08/02/introducing-workspaces/). 

Currently, the bundled file is very huge (4 MB when used with `Yarn Workspaces`) compared to the same project when not used with `Yarn Workspaces` (25 kB). `serverless-webpack` bundles all the dependencies in one file even when using `webpack-node-externals` and `webpackIncludeModules`. 

Thanks! 😃

HyperBrain:
## What did you implement:

Closes #286 

## How did you implement it:

Yarn support is provided by an additional packager class that uses the yarn executable instead of npm.
The packager is selected by the `webpack: packager` setting introduced in serverless-webpack version 5.0.0.

The implementation honors any existing yarn.lock file and provides the same reproducibility as the locked npm packages. It even should behave better, because `--frozen-lockfile` is used with the `yarn install` command. Additionally, some special packager options are supported in `webpack: packagerOptions`:
```yaml
custom:
  webpack:
    webpackConfig: ./webpack.config.js
    includeModules: true
    packager: yarn  # has to be set explicitly as it defaults to npm
    packagerOptions:
      flat: true   # Will set --flat on install (defaults to false)
      ignoreScripts: true  # Do not execute package.json hook scripts on install (defaults to false)
```

## How can we verify it:

Use Yarn in your project - preferably with a yarn.lock file - and set the packager accordingly in `serverless.yml`. A `serverless package` or `serverless deploy` should do as before, but use yarn and the yarn.lock to create the packages.
You can verify the installation by inspecting the generated zip files in `.serverless` (best to be done with `serverless package`.

With Yarn it is not needed to copy the modules explicitly into the function dependencies. For npm copy+prune was faster than install, but yarn is fast as hell here!
**With that optimization, the yarn based package/deployment should take less than half of the time that npm does !!! Of course this heavily depends on the machine where it is executed. With a very large file system cache and big amounts of memory, the difference might be rather small.**

## Your help needed 😄 

There are still some uncertainties, especially, if transient dependencies are handled correctly (i.e. if you bundle a first level dependency, its second level dependencies must become first level dependencies). With npm this works as expected and the behavior with yarn should be exactly the same.
Second, we should try this with the most common installation flavors of yarn, from a simple project up to yarn workspaces and shared modules.

## Todos:

Unit tests and proper documentation in the README are yet missing. Working on that.

- [x] Write tests
- [x] Write documentation
- [x] Fix linting errors

### DIFF
--- a/README.md
+++ b/README.md
@@ -276,7 +276,7 @@ Right now there are no `packagerOptions` that can be set with NPM.
 
 #### Yarn
 
-Using yarn will switch the whole packaging pipeline to use yarn, so does it use a `yan.lock` file.
+Using yarn will switch the whole packaging pipeline to use yarn, so does it use a `yarn.lock` file.
 
 The yarn packager supports the following `packagerOptions`:
 
--- a/README.md
+++ b/README.md
@@ -282,7 +282,6 @@ The yarn packager supports the following `packagerOptions`:
 
 | Option        | Type | Default | Description |
 |---------------|------|---------|-------------|
-| flat          | bool | true    | Will set --flat on install |
 | ignoreScripts | bool | true    | Do not execute package.json hook scripts on install |
 
 #### Forced inclusion
--- a/README.md
+++ b/README.md
@@ -246,11 +246,44 @@ custom:
 ```
 > Note that only relative path is supported at the moment.
 
-#### Usage with yarn
+#### Packagers
 
-Note that if auto-packing is enabled, the plugin will call `npm install`. If you are using yarn your `yarn.lock` file will be not be honored, which might lead to unexpected results as your dependencies will most likely not match (or be missing, as npm does not install packages in the same way as yarn).
+You can select the packager that will be used to package your external modules.
+The packager can be set with the packager configuration. Currently it can be 'npm'
+or 'yarn' and defaults to using npm when not set.
 
-Yarn support is planned [#286][link-286]. Until then we recommend using npm when using auto-packing.
+```yaml
+# serverless.yml
+custom:
+  webpack:
+    packager: 'yarn'      # Defaults to npm
+    packagerOptions: {}   # Optional, depending on the selected packager
+```
+
+You should select the packager, that you use to develop your projects, because only
+then locked versions will be handled correctly, i.e. the plugin uses the generated 
+(and usually committed) package lock file that is created by your favorite packager.
+
+Each packager might support specific options that can be set in the `packagerOptions`
+configuration setting. For details see below.
+
+##### NPM
+
+By default, the plugin uses NPM to package the external modules. However, if you use npm,
+you should use any version `<5.5 >=5.7.1` as the versions in-between have some nasty bugs.
+
+Right now there are no `packagerOptions` that can be set with NPM.
+
+#### Yarn
+
+Using yarn will switch the whole packaging pipeline to use yarn, so does it use a `yan.lock` file.
+
+The yarn packager supports the following `packagerOptions`:
+
+| Option        | Type | Default | Description |
+|---------------|------|---------|-------------|
+| flat          | bool | true    | Will set --flat on install |
+| ignoreScripts | bool | true    | Do not execute package.json hook scripts on install |
 
 #### Forced inclusion
 
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -1,6 +1,5 @@
 environment:
   matrix:
-    - NODEJS_VERSION: "4"
     - NODEJS_VERSION: "6"
 
 install:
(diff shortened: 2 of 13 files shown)


HyperBrain:
@franciscocpg This is the promised Yarn support. Can you give it a try and tell me, if it works for you?

RishitKedia:
@HyperBrain Hey, Frank! 👋

Thanks for your time and work on this! You're awesome. ❤️

Just tested this on my project, and I can confirm that it works! 💪

franciscocpg:
@HyperBrain 
I'll try ASAP

j0k3r:
❓  _I don't know if I have to submit a different issue. If you prefer, I can remove my comment and create a separate issue._

But I just tried on a project and got an error because of flatten deps.

In my `package.json` I have `"colors": "^1.1.2",` and in the `yarn.lock` I have:

```
colors@1.0.x:
  version "1.0.3"
  resolved "https://registry.yarnpkg.com/colors/-/colors-1.0.3.tgz#0433f44d809680fdeb60ed260f1b0c262e82a40b"

colors@^1.1.2:
  version "1.1.2"
  resolved "https://registry.yarnpkg.com/colors/-/colors-1.1.2.tgz#168a4701756b6a7f51a12ce0c97bfa28c084ed63"
```

Because winston requires the version 1.0.x:
```
winston@^2.4.0:
  version "2.4.0"
  resolved "https://registry.yarnpkg.com/winston/-/winston-2.4.0.tgz#808050b93d52661ed9fb6c26b3f0c826708b0aee"
  dependencies:
    async "~1.0.0"
    colors "1.0.x"
    cycle "1.0.x"
    eyes "0.1.x"
    isstream "0.1.x"
    stack-trace "0.0.x"
```
When running `yarn install --frozen-lockfile --non-interactive --flat --ignore-scripts` I got:

```
$ yarn install --frozen-lockfile --non-interactive --flat --ignore-scripts
yarn install v1.5.1
[1/4] 🔍  Resolving packages...
info Unable to find a suitable version for "colors", please choose one by typing one of the numbers below:
  1) "colors@^1.1.2" which resolved to "1.1.2"
  2) "colors@1.0.x" which resolved to "1.0.3"
Answer?:
```

Which means during the `serverless package` I got that error:

>   Error --------------------------------------------------
>
>  Command failed: yarn install --frozen-lockfile --non-interactive --flat --ignore-scripts
error An unexpected error occurred: "Can't answer a question unless a user TTY".

You do have `--non-interactive` so I don't know how can this be fixed.

Using only `yarn install` doesn't generate that prompt. The prompt arrive when I use `--flat`.
Which means when flatten deps it needs to chose only one version 😕

HyperBrain:
@j0k3r I'm not sure if that should be the default behavior when flattening as the resolution cannot be determined. We have to use non-interactive mode during packaging.

Do you know if there's a specific resolution switch for yarn, where I can tell to use the latest version, or provide a resolution strategy to yarn?

j0k3r:
@HyperBrain That could be a good option but I've no idea of such option (and I'm not a yarn expert)

HyperBrain:
Ok. Here's a feature request at Yarn: https://github.com/yarnpkg/yarn/issues/1658

It seems that automatic resolution is not yet supported for the flat install option and it is not possible to resolve in non-interactive runs. We should then postpone the flat option support until a solution is available on the Yarn end. Right now I do not see any way to use that in the plugin.

README.md
```diff
@@ -246,11 +246,44 @@ custom:
...
+##### NPM
+
+By default, the plugin uses NPM to package the external modules. However, if you use npm,
+you should use any version `<5.5 >=5.7.1` as the versions in-between have some nasty bugs.
+
+Right now there are no `packagerOptions` that can be set with NPM.
+
+#### Yarn
+
+Using yarn will switch the whole packaging pipeline to use yarn, so does it use a `yan.lock` file.
```
Gerst20051: should be `yarn.lock` not `yan.lock`

HyperBrain: Good catch. Thanks. 👍

j0k3r:
I'll give an other try tomorrow!

SebastianEdwards:
@HyperBrain another Yarn success story here. Works like a charm 👍

j0k3r:
@HyperBrain Works like a charm too.
Really good job 👏

HyperBrain:
Merged.

PR #351, opened 2018-03-20
Purge handler module from require cache on watch

### LINKED ISSUE #349: invoke local --watch is not working
# This is a Bug Report

## Description

* What went wrong?
start with
`sls invoke local --function some-function --watch`
when change the code, it recompiles the code,  but it still seems invoke the old compiled result

It should run the latest compiled result just as what the documentation says.

```
service: some-service

plugins:
  - serverless-webpack
  - serverless-offline

custom:
  webpack:
    webpackIncludeModules: true

package:
  individually: true

provider:
  name: aws
  runtime: nodejs6.10
  stage: dev
  region: us-east-2

functions:
  some-function:
    handler: some-function/handler.handler
```

No errors, the output is like this
image

the second time it should output 5 instead of 6 because the code has changed.

## Additional Data

* ***Serverless-Webpack Version you're using***:  5.1.0
* ***Webpack version you're using***: 4.1.1
* ***Serverless Framework Version you're using***: 1.26.1
* ***Operating System***: macOs High Sierra  10.13.3

HyperBrain:
## What did you implement:

Closes #349 

## How did you implement it:

Before Serverless requires the handler, we purge any loaded ones from the require cache.
This forces Serverless to load a fresh handler from file, which is needed in case a watch compile happened. Otherwise the old module from the initial compile will be used.

## How can we verify it:

* Invoke any function locally in watch mode `npx sls invoke local --function=XXXX --path=some-event.json --watch`
* Check the output
* Change any functionality in the sources and save them to disk
* Wait for the recompile and invocation thereafter
* Check the output. Any changes in the sources should be reflected.

## Todos:

- [x] Write tests
- ~~Write documentation~~ (N/A)
- [x] Fix linting errors

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -34,15 +34,16 @@ function searchAndProcessCache(moduleName, processor) {
   let mod_src = require.resolve(moduleName);
   const visitedModules = [];
   if (mod_src && ((mod_src = require.cache[mod_src]) !== undefined)) {
-    (function traverse(mod) {
+    const modStack = [mod_src];
+
+    while (!_.isEmpty(modStack)) {
+      const mod = modStack.pop();
       if (!_.includes(visitedModules, mod)) {
         visitedModules.push(mod);
-        _.forEach(mod.children, function (child) {
-          traverse(child);
-        });
+        Array.prototype.push.apply(modStack, mod.children);
         processor(mod);
       }
-    }(mod_src));
+    }
   }
   return BbPromise.resolve();
 }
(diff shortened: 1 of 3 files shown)


lib/utils.js
```diff
@@ -11,31 +12,43 @@ function guid() {
...
     (function traverse(mod) {
-      _.forEach(mod.children, function (child) {
-        traverse(child);
-      });
-      callback(mod);
-    }(mod));
+      if (!_.includes(visitedModules, mod)) {
+        visitedModules.push(mod);
+        _.forEach(mod.children, function (child) {
+          traverse(child);
```
HyperBrain: To prevent stackoverflows I'll replace the recursive implementation with a sequential one (breadth or depth first does not matter here).

PR #373, opened 2018-04-25
Do not swallow errors in case npm ls fails without an empty stdout and use spawn for child processes

HyperBrain:
## What did you implement:

If `npm ls` fails the plugin tries to analyze stderr for npm errors that can be ignored. If none are found, it will pass stdout to `JSON.parse()`.
The case that any non-npm error was happening was not handled correctly (e.g. if the stdout maxbuffer size got exceeded by a very big dependency section). The plugin passed the empty stdout to JSON parse and failed there without telling the exact reason for the failure.
Additionally, Node's buffered stdio streams do not work reliably and the `exec` method is not meant to be used for processes that might return data of arbitrary length.

## How did you implement it:

The packager executables are now executed with `child_process.spawn` which allows for stdio streams of arbitrary length. On execution failure, a proper SpawnError is now thrown that contains the full stdout and stderr streams.
Additionally this eliminates the need of the `packExternalModulesMaxBuffer` buffer configuration which has been removed from the README.
The implementation should seamlessly work as before.

## How can we verify it:

Use a package.json with _many_ dependencies, so that the output of `npm ls` exceeds the max stdout buffer default of 200k with the old serverless-webpack version.
It should work with this branch now.

## Todos:

- [x] Write tests
- [x] Write documentation
- [x] Fix linting errors

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,7 +61,6 @@ custom:
     webpackConfig: 'webpack.config.js'   # Name of webpack configuration file
     includeModules: false   # Node modules configuration for packaging
     packager: 'npm'   # Packager that will be used to package your external modules
-    packExternalModulesMaxBuffer: 200 * 1024   # Size of stdio buffers for spawned child processes
 ```
 
 ### Webpack configuration file
--- a/lib/Configuration.js
+++ b/lib/Configuration.js
@@ -13,7 +13,6 @@ const DefaultConfig = {
   includeModules: false,
   packager: 'npm',
   packagerOptions: {},
-  packExternalModulesMaxBuffer: 200 * 1024,
   config: null
 };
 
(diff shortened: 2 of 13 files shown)


HyperBrain:
Hi @franciscocpg . Could you do me a favor and test if the current master still works properly for you? I switched the packager invocations to use `child_process.spawn` instead of exec here, but only verified Windows yet. It would be great if I had an additional success report before I release the fix 😃 . Thanks.

franciscocpg:
hi @HyperBrain 
I'll do it.

A `sls package` would be enough?

HyperBrain:
Thanks 😏 . Yes I think so, because it runs through the whole npm or yarn processes. There's no need to actually deploy it.

Issue #383, opened 2018-05-02
AWS error: Unable to import module 'src/service/serviceName/handler': Error


jascination:
# This is a Bug Report

## Description

This vague issue is mentioned in other bug reports, so apologies if it's a duplicate, but while the error code is the same those issues/resolutions seem unrelated to what's happening to me.

* What went wrong?

When deploying to AWS and trying to access my lambdas, I'm getting:

```
Unable to import module 'src/service/authentication/handler': Error
```

This is the only error I see (no stacktrace...).

* What was the config you used?
serverless.yml:

```
service: ${opt:service, 'OurApplicationName'}

frameworkVersion: '>=1.0.0 <2.0.0'

provider:
  name: aws
  runtime: nodejs8.10
  stage: ${opt:stage, 'dev'}

  region: 'ap-southeast-2'
  role: '${self:custom.roles.default}'

  deploymentBucket:
    name: '${self:provider.stage}-${self:service}-deployment-bucket'
    serverSideEncryption: 'AES256'

  # VPC configuration per environment (dev, qa, prod)
  # This is required for deployment, but not during Lambda execution.
  vpc: '${file(./serverless-vpc-config.yml):${opt:stage, self:provider.stage}}'

  environment: '${file(.env.${opt:stage, self:provider.stage}.yml)}'

  iamRoleStatements:
    - Effect: 'Allow'
      Action:
        - 'lambda:InvokeFunction'
      Resource:
      - Fn::Join:
        - ':'
        - - arn:aws:lambda
          - Ref: AWS::Region
          - Ref: AWS::AccountId
          - function:${self:service}-${opt:stage, self:provider.stage}-*


functions:
  ${file(resource/serverlessHelper.js):aggregatedFunctions}

plugins:
  - serverless-swagger-documentation
  - serverless-webpack
  - serverless-dynamodb-local
  - serverless-s3-local
  - serverless-offline
  - serverless-offline-scheduler
  - serverless-domain-manager

resources:
  Resources:
    '${file(resource/serverlessHelper.js):cloudformationResources}'

custom:
   # A bunch of dynamoDB seeding stuff here


  # serverless-webpack configuration
  webpack:
    includeModules: true

 ```

My folder structure is like so:

```
/
├── resource
│   └── // database initialization, swagger documentation tooling, authentication...
├── src
│   ├── service
│   │   └── profile
│   │       ├── handler.js
│   │       ├── profile.fn.yml
       └── // All services have their own folders in here
[...]

Issue #369, opened 2018-04-14
Error when packager is yarn


bazaglia:
# This is a Bug Report

## Description

When using nodeExternals in webpack config, deployment throws an error. The issue is only reproduced with `includeModules: true` + `packager: 'yarn'`.

```
 Error --------------------------------------------------
 
  Command failed: yarn install --frozen-lockfile --non-interactive
error An unexpected error occurred: "Your lockfile needs to be updated, but yarn was run with `--frozen-lockfile`.".

```

It may help because the deployment only include production packages, which causes a difference in yarn lockfile.

## Additional Data

* ***Serverless-Webpack***: 5.1.1
* ***Webpack version***: 4.5.0
* ***Serverless Framework***: 1.26.1
* ***webpack-node-externals***: 1.7.2

mzmiric5:
I've also noticed the same behavior, but only when I use local packages with the file: reference in my package.json

HyperBrain:
@bazaglia @mzmiric5 This is intended behavior. The plugin sets `--frozen-lockfile` to prevent the build/deployment server install any new versions on its own and so change the module dependencies.

You should check, if your yarn.lock file is up-to-date and matches everything that is in your package.json. If you change your package.json manually, they are out of sync and the packaging will fail. 

As long as you do a `yarn install` after changing the files and use `yarn add` to add linked of local (file) packages, it will generate a matching yarn.lock file and should work.

However, I can add an option like `custom: webpack: packagerOptions: noFrozenLockFile: true` that can be set in the `serverless.yml`. Imo it must default to freeze because of CI/CD stability but then you could disable the behavior easily for unstable testing environments.

mzmiric5:
@HyperBrain oh yeah, I mean I totally agree that frozen lockfile should be a feature that is used for build stability, and in my case, if I just run the exact same yarn command outside of the serverless build process, I don't get this error. I guess I can make a separate issue regarding this since it seems to be affecting npm as well for me (which I was about to do anyway but I saw this issue, and it was describing the same final output I was getting so I tagged along).

HyperBrain:
> I guess I can make a separate issue regarding this since it seems to be affecting npm as well

Yes, please do that. Then the file reference/lock issue can be solved separately.

bazaglia:
[...]

PR #384, opened 2018-05-02
Added error if needed runtime dependency is a devDependency

HyperBrain:
## What did you implement:

Relates to #383, https://github.com/serverless-heaven/serverless-webpack/issues/369#issuecomment-385783247, #230

This PR introduces better error handling for the case that a runtime dependency detected by Webpack is only found in `devDependencies` and would lead to an `Unable to import module` error when deployed.
If the detected module is forcefully excluded it will, however, succeed.

## How did you implement it:

Added a check for devDependencies in the production module loop.

## How can we verify it:

Move a needed dependency to devDependencies. It should error out as long as the dependency is not forcefully excluded.

## Todos:

- [x] Write tests
- ~~Write documentation~~
- [x] Fix linting errors


HyperBrain:
Released with 5.1.5

PR #391, opened 2018-05-12
added allow exported unnamed func

### LINKED ISSUE #372: Missing entries in slsw.lib when serverless functions are defined in a separate file
# This is a Bug Report

## Description

For bug reports:
* What went wrong?

Webpack fails to identify entry point when serverless functions are extracted out into a separate yml file. However, it works when serverless.yml has the functions defined in the same file.

Should have identified the correct entry points in `slsw.lib.entries`

❗️  **This configuration fails**

*serverless.yml*

```yml
functions:
    - ${file(src/functions/index.yml)}
```

*functions/index.yml*

```yml
index:
  handler: src/handler.handle
  events:
    - http
```

**This configuration works fine**

*serverless.yml*

```yml
functions:
  misc:
    handler: src/handler.handle
    events:
      - http
```

```

  Webpack Options Validation Error -----------------------

  Invalid configuration object. Webpack has been initialised using a configuration object that does not match the API schema.
 - configuration.entry should be one of these:
   object { <key>: non-empty string | [non-empty string] } | non-empty string | [non-empty string] | function
   -> The entry point(s) of the compilation.
   Details:
    * configuration.entry should not be empty.
      -> Multiple entry bundles are created. The key is the chunk name. The value can be a string or an array.
    * configuration.entry should be a string.
      -> An entry point without name. The string is resolved to a module which is loaded upon startup.
    * configuration.entry should be an array:
      [non-empty string]
    * configuration.entry should be an instance of function
      -> A Function returning an entry object, an entry string, an entry array or a promise to these things.
```

## Additional Data

* ***Serverless-Webpack Version you're using***: 5.1.0
* ***Webpack version you're using***: 4.2.0
* ***Serverless Framework Version you're using***: 1.26.1
* ***Operating System***: linux

immexerxez:
## What did you implement:

Closes #372 

## How did you implement it:

quick fix to allow unnamed funcs to be packed w/ lib.entries instead of having to specify each one in webpack config

## How can we verify it:

try to package a serverless project w/ an unnamed function

## Todos:

- [ ] Write tests
- [ ] Write documentation
- [ ] Fix linting errors

### DIFF
--- a/lib/validate.js
+++ b/lib/validate.js
@@ -24,7 +24,7 @@ module.exports = {
   validate() {
     const getHandlerFile = handler => {
       // Check if handler is a well-formed path based handler.
-      const handlerEntry = /(.*)\..*?$/.exec(handler);
+      const handlerEntry = /(.*)\..*?$/.exec(handler) || /(.*).*?$/.exec(handler);
       if (handlerEntry) {
         return handlerEntry[1];
       }


immexerxez:
Closing because it conflicts with lambda required export named function

PR #413, opened 2018-06-20
Added support for async webpack config object

### LINKED ISSUE #412: Asynchronous configuration object support for Webpack
# This is a  Feature Proposal

## Description
Webpack 2+ support async configuration object by exporting a promise as configuration object.
This is a feature request to support async configuration object. This feature helps to define env variables and constants inside code after an async operation inside `webpack.config.js`. for example defining `AccountId` of current user inside `lambda@edge` which does not support defining normal process environment variables.

## Additional Data

* ***Serverless-Webpack Version***: 5.1.5
* ***Webpack version***: 4.11.1
* ***Serverless Framework Version***: 1.27.2
* ***Operating System***: MacOS High Sierra (10.13.5)
* ***Stack Trace***:

nidin:
## What did you implement:

Closes #412

Webpack 2+ support async configuration object by exporting a promise as configuration object.
This PR added support for async configuration object.

## How did you implement it:

` Line 114 :  this.webpackConfig = require(webpackConfigFilePath);`
Return an object or a promise depending on user's webpack config.
All the validation code after this logic moved in to an arrow function called `processConfig`. 
First argument of this function is `_config` and it will overwrite the variable `this.webpackConfig`.
After defining the function, an if condition added to check whether the config object is a promise or a normal object. If it is a promise `processConfig` will call with resolved value otherwise immediately call `processConfig`. In both case `processConfig` will get webpack config object as it's first argument.

## How can we verify it:

Wrap webpack config in a Promise

Example:
```
const path = require('path');
const webpack = require('webpack')
const AWS = require('aws-sdk');

const outDir = '.build/.webpack';

module.exports = new Promise(function (resolve) {
    slsw.lib.serverless.providers.aws.getAccountId().then(accountId => {
        resolve({
            entry: slsw.lib.entries,
            target: 'node',
            mode: slsw.lib.webpack.isLocal ? 'development' : 'production',
            plugins: [
                new webpack.DefinePlugin({
                    AWS_ACCOUNT_ID: `${accountId}`,
                }),
            ],
            output: {
                libraryTarget: 'commonjs',
                path: path.resolve(__dirname, outDir),
                filename: '[name].js',
                devtoolModuleFilenameTemplate: function (info) {
                    return "file:///" + encodeURI(info.absoluteResourcePath);
                },
            },
        });
    })
});
```

## Todos:

- [x] Write tests
- [x] Write documentation
- [x] Fix linting errors

### DIFF
--- a/lib/validate.js
+++ b/lib/validate.js
@@ -118,103 +118,112 @@ module.exports = {
       }
     }
 
-    // Default context
-    if (!this.webpackConfig.context) {
-      this.webpackConfig.context = this.serverless.config.servicePath;
-    }
-
-    // Default target
-    if (!this.webpackConfig.target) {
-      this.webpackConfig.target = 'node';
-    }
-
-    // Default output
-    if (!this.webpackConfig.output || _.isEmpty(this.webpackConfig.output)) {
-      const outputPath = path.join(this.serverless.config.servicePath, '.webpack');
-      this.webpackConfig.output = {
-        libraryTarget: 'commonjs',
-        path: outputPath,
-        filename: '[name].js',
-      };
-    }
-
-    // Custom output path
-    if (this.options.out) {
-      this.webpackConfig.output.path = path.join(this.serverless.config.servicePath, this.options.out);
-    }
-
-    if (this.skipCompile) {
-      this.serverless.cli.log('Skipping build and using existing compiled output');
-      if (!fse.pathExistsSync(this.webpackConfig.output.path)) {
-        return BbPromise.reject(new this.serverless.classes
-          .Error('No compiled output found'));
-      }
-      this.keepOutputDirectory = true;
-    }
-
-    if (!this.keepOutputDirectory) {
-      this.options.verbose && this.serverless.cli.log(`Removing ${this.webpackConfig.output.path}`);
-      fse.removeSync(this.webpackConfig.output.path);
-    }
-    this.webpackOutputPath = this.webpackConfig.output.path;
-
-    // In case of individual packaging we have to create a separate config for each function
-    if (_.has(this.serverless, 'service.package') && this.serverless.service.package.individually) {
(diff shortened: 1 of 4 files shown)


lib/validate.js
```diff
@@ -117,104 +117,121 @@ module.exports = {
...
-          [entryFunc.entry.key]: entryFunc.entry.value
-        };
-        const compileName = entryFunc.funcName || _.camelCase(entryFunc.entry.key);
-        config.output.path = path.join(config.output.path, compileName);
-        return config;
+      return BbPromise.resolve();
+    };
+    
+    // Webpack config can be a Promise, If it's a Promise wait for resolved config object.
+    if (this.webpackConfig && this.webpackConfig instanceof Promise) {
```
HyperBrain: This check most likely will fail. The plugin as well as Serverless uses bluebird promises, and so should this method. We have to check for a promise by checking if it is a thenable, not a specific Promise type.

lib/validate.js
```diff
@@ -117,104 +117,121 @@ module.exports = {
...
+      return new BbPromise((resolve, reject) => {
```
HyperBrain: We should remove the `new Promise` anti-pattern here.
```
return this.webpackConfig.then(processConfig);
```

lib/validate.js
nidin: That's true. I will change it by checking thenable.

lib/validate.js
nidin: Copied.

nidin:
@HyperBrain I was wondering in which section of the README should I add the new functionality?

HyperBrain:
It would fit after the `A base Webpack configuration might look like this:` example. We could just mention that a promised config can be used alternatively, with a small example.

nidin:
Copied 👍🏻

nidin:
@HyperBrain Requested changes are committed. You may review it again. Feel free to remove any unwanted writings from README.

PR #450, opened 2018-09-26
commonjs2 is the one that nodejs is using

stevemao:
See https://github.com/webpack/webpack/issues/1114

***Is it a breaking change?:*** NOT for most users. since this plugin is for nodejs only

### DIFF
--- a/README.md
+++ b/README.md
@@ -227,7 +227,7 @@ const path = require('path');
 module.exports = {
   // ...
   output: {
-    libraryTarget: 'commonjs',
+    libraryTarget: 'commonjs2',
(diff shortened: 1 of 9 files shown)


stevemao:
I also think that all examples should use `commonjs2`. There's actually no environment that's using `commonjs`

HyperBrain:
Hi @stevemao ,
thanks for the PR. I think you are right, and we should be safe to change the default to `commonjs2`.
Can you also adapt the unit tests, because they are now failing (they still check for `commonjs`).
... and feel free to change the examples in this PR too

PR #451, opened 2018-09-26
includeModules should be true by default

stevemao:
If a module is not bundled, it will automatically package it.
If all modules are bundled, it will package nothing extra. So this is more intuitive.

***Is it a breaking change?:*** NOT for most users. I can't why if a module is not bunded, one doesn't want to package it.

### DIFF
--- a/lib/Configuration.js
+++ b/lib/Configuration.js
@@ -10,7 +10,7 @@ const _ = require('lodash');
  */
 const DefaultConfig = {
   webpackConfig: 'webpack.config.js',
-  includeModules: false,
+  includeModules: true,
   packager: 'npm',
   packagerOptions: {},
   config: null


designfrontier:
@stevemao After some thought and discussion we are going to decline your PR. This is certainly not the most intuitive change, but it would require all of our users to check their config and ensure that no change was needed for their project.

We're not opposed to breaking changes, but this particular change would be difficult to communicate out and it is a subtle enough change that people would certainly end up broken because of it.

Is there documentation that we could add that would make the current state of the setting clearer in your mind?

Thank you for your willingness to help out!

stevemao:
I believe most people already use:

```yml
# serverless.yml
custom:
  webpack:
    includeModules:
      forceExclude:
        - aws-sdk
```

So this option should impact minimal users. 

Also this change will potentially pack packages that users "forgot" to pack. So their code should automatically work and it's more like a bug fix than breaking change.

stevemao:
Looking at it again, once you do:

```
# serverless.yml
custom:
  webpack:
    includeModules:
      forceExclude:
        - aws-sdk
```

if `includeModules` is `false` by default, the above code should do nothing since it's already not packing any deps. But if it's `true` then I'm consciously force exclude my `aws-sdk`.

PR #458, opened 2018-10-25
Update get version for external transient dependancies for yarn

peterlenagh:
## What did you implement:

Hi, I don't think the transient dependancy versions are being looked up correctly if you use yarn.

It looks in `dependencyGraph[module.origin].dependencies[module.external].version` but for yarn I think it would just be `dependencyGraph[module.external].version`?

## How did you implement it:

check packager and if yarn use `dependencyGraph[module.external].version` in moduleVersion bit of getProdModules.

This worked for me™ but idk if it's the best way to do it, happy to update this PR if there's a better way to approach this.

## How can we verify it:

My use case was as follows - 

using serverless-webpack with yarn, including a private npm module that needs to be transpiled/bundled (it contains universal js for SSR).

also using nodeExternals and `includeModules:true`. so that private npm module is whitelisted in nodeExternals, and included in the babel rule in webpack.

before the change in this PR, on running `sls webpack` I would get warnings `WARNING: Could not determine version of XX` where XX is any of the deps in my private module (not directly in the ssr app package.json) and then it would error ```yarn install --frozen-lockfile --non-interactive failed with code 1
error Your lockfile needs to be updated, but yarn was run with `--frozen-lockfile```

with the change, it finds the nested dep versions and builds correctly.

## Todos:

- [ ] Write tests
- [ ] Write documentation
- [ ] Fix linting errors

***Is it a breaking change?:*** NO?

### DIFF
--- a/lib/packExternalModules.js
+++ b/lib/packExternalModules.js
@@ -98,8 +98,12 @@ function getProdModules(externalModules, packagePath, dependencyGraph, forceExcl
     } else {
       if (!packageJson.devDependencies || !packageJson.devDependencies[module.external]) {
         // Add transient dependencies if they appear not in the service's dev dependencies
-        const originInfo = _.get(dependencyGraph, 'dependencies', {})[module.origin] || {};
-        moduleVersion = _.get(_.get(originInfo, 'dependencies', {})[module.external], 'version');
+        if (this.configuration.packager !== 'yarn') {
+          const originInfo = _.get(dependencyGraph, 'dependencies', {})[module.origin] || {};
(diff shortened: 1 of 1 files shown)


peterlenagh:
actually this doesnt even fix it, i still get 
```yarn install --frozen-lockfile --non-interactive failed with code 1
error Your lockfile needs to be updated, but yarn was run with `--frozen-lockfile`.```

maybe yarn install shouldnt be using frozen-lockfile here?

peterlenagh:
OK after looking into this more i think this is correct to be using --frozen-lockfile, but it should also be able to work out the versions of these deps. Closing this PR because its invalid and I've opened an issue : https://github.com/serverless-heaven/serverless-webpack/issues/460

PR #462, opened 2018-11-11
Speed up cleanup process

neilime:
## What did you implement:

Removing ".webpack" directory may be is very slow on when this directory is huge.

## How did you implement it:

- Performs remove async (no need to wait the remove of this directory before running  the next process)

## How can we verify it:

Run serverless deploy

## Todos:

- [x] Write tests
- [x] Write documentation
- [x] Fix linting errors

### DIFF
--- a/lib/cleanup.js
+++ b/lib/cleanup.js
@@ -1,20 +1,26 @@
 'use strict';
 
+const _ = require('lodash');
 const BbPromise = require('bluebird');
 const fse = require('fs-extra');
 
 module.exports = {
   cleanup() {
     const webpackOutputPath = this.webpackOutputPath;
-
     const keepOutputDirectory = this.keepOutputDirectory;
+    const cli = this.options.verbose ? this.serverless.cli : { log: _.noop };
+
     if (!keepOutputDirectory) {
-      this.options.verbose && this.serverless.cli.log(`Remove ${webpackOutputPath}`);
+      cli.log(`Remove ${webpackOutputPath}`);
       if (this.serverless.utils.dirExistsSync(webpackOutputPath)) {
-        fse.removeSync(webpackOutputPath);
+        // Remove async to speed up process
+        fse
+          .remove(webpackOutputPath)
+          .then(() => cli.log(`Removing ${webpackOutputPath} done`))
+          .catch(error => cli.log(`Error occurred while removing ${webpackOutputPath}: ${error}`));
       }
     } else {
-      this.options.verbose && this.serverless.cli.log(`Keeping ${webpackOutputPath}`);
+      cli.log(`Keeping ${webpackOutputPath}`);
     }
 
     return BbPromise.resolve();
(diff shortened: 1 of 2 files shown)


neilime:
Any chance to get this PR reviewed ?

HyperBrain:
@serverless-heaven/serverless-webpack-contributors Can someone help to review this PR? If it is ok, it will be part of 5.3.0

designfrontier:
I'll take a look at it right now...

lib/cleanup.js
```diff
@@ -4,15 +4,23 @@ const BbPromise = require('bluebird');
 const fse = require('fs-extra');
 
 module.exports = {
-  cleanup() {
-    const webpackOutputPath = this.webpackOutputPath;
+	cleanup() {
+		const webpackOutputPath = this.webpackOutputPath;
+		if (this.serverless.utils.dirExistsSync(webpackOutputPath)) {
+      
+      const cli = this.options.verbose ? this.serverless.cli : null;
```
designfrontier: Rather then returning null and having to check everywhere before logging it might be easier to return an object with a void function here. Something like `{ log: () => {} }`. That would let you just call `cli.log` without the check making the subsequent code a little less verbose.

lib/cleanup.js
```diff
@@ -4,15 +4,23 @@ const BbPromise = require('bluebird');
 const fse = require('fs-extra');
 
 module.exports = {
-  cleanup() {
-    const webpackOutputPath = this.webpackOutputPath;
+	cleanup() {
```
designfrontier: Looks like the code style calls for 2 spaces as indent here

neilime:
@designfrontier : I've made some change tell me if it's good to you now, thank you.

lib/cleanup.js
designfrontier: should be a second space here... other than that it looks fantastic

lib/cleanup.js
```diff
@@ -4,19 +4,26 @@ const BbPromise = require('bluebird');
...
       }
     } else {
-      this.options.verbose && this.serverless.cli.log(`Keeping ${webpackOutputPath}`);
+      cli.log(`Keeping ${webpackOutputPath}`);
     }
 
-    return BbPromise.resolve();
-  },
+		return BbPromise.resolve();
+	},
```
designfrontier: this should have a second space as well

designfrontier:
one super tiny thing. Thanks for your work on this

HyperBrain:
@designfrontier Can you finish the review and eventually merge?

designfrontier:
Yeah I’ll take a look today

hassankhan:
Just had a chance to test it on my own projects, seems to work fine but can't really tell since I don't have large deployment sizes.

designfrontier:
Actually... on a second look it looks like there is a new branch in the code that is untested. Details are here: https://coveralls.io/builds/23015127/source?filename=lib/cleanup.js

I think you should be able to test this by setting everything up and removing the webpack directory before you run cleanup. Then checking that you get the correct error output in the log with verbose set.

Bonus points would be testing that ternary that changes where the logs go

PR #474, opened 2019-01-12
Fix behavior of file rebasing when file starts with "file:./"

talbenari1:
Relative files that are located in the same directory as the `package.json`, e.g. `"file:./foo.tgz"`, were not being rebased correctly by `rebaseFileReferences` due to a bad regular expression. The regex has
been fixed in this commit.


lib/packExternalModules.js
```diff
@@ -9,7 +9,7 @@ const isBuiltinModule = require('is-builtin-module');
 const Packagers = require('./packagers');
 
 function rebaseFileReferences(pathToPackageRoot, moduleVersion) {
-  if (/^(?:file:[^/]{2}|\.\/|\.\.\/)/.test(moduleVersion)) {
+  if (/^(file:[^\/]|\.\/|\.\.\/)/.test(moduleVersion)) {
```
HyperBrain: You should keep the `?:` as it marks the group as non-capturing group here.

HyperBrain: Additionally, I do not know if a `file:` prefix is a valid file URL. Normally only `file://` is valid, but relative files, as in your example, should be written as `file://./` or without the file prefix?

talbenari1: Sure thing, I can put the non-capturing group specifier back.

Regarding the prefix, this is the function that handles package resolution in `npm`: https://github.com/npm/npm-package-arg/blob/latest/npa.js#L51-L79. As you can see, the prefix it uses is simply `file:` rather than `file://`.

talbenari1: In fact, a much more correct solution here would be to pull in `npm-package-arg` as a dependency, but I didn't want to add a new dependency on your behalf.

HyperBrain: @talbenari1 For now we can go with the light-weight regex approach. 

However, I like the idea going with `npm-package-arg` but think this would be more like a overhaul of different parts of the dependency management within the plugin - at least to make it consistent. If you want, you can open a new issue and put the use of the module under discussion there. I'm curious what others will think about the idea and we might get a broader picture and decide how to proceed.

hassankhan: Agree with @HyperBrain, let's keep the regex approach for this PR, but look into using `npm-package-arg` going forward. I'm all for moving complexity out of the plugin 😄

miguel-a-calles-mba: Has this comment been addressed?

miguel-a-calles-mba:
```suggestion
  if (/^(file:[^/]|\.\/|\.\.\/)/.test(moduleVersion)) {
```

PR #485, opened 2019-02-19
ability to not follow symlinks when packaging

kevinsperrine:
## What did you implement:
I needed this, because I built a custom binary of imagemagick to use in lambda, and it contains lots of symlinks. When these symlinks were dereferenced the filesize was 280MB and therefore too large for lambda. I added a `symlink: false` (default is true) flag to individual functions, and allows me to choose whether we want to include dereferenced symlinks or not. Related to https://github.com/serverless/serverless/issues/3215, and I'll probably work on porting it over there, but implemented here first since I use webpack for all of my deploys.

## How did you implement it:
I check if the symlink flag exists and pass an additional parameter to the zip function in packageModules. If we're not following symlinks, then I used lstatSync instead of statSync to get the details about the file, so that we can know if it's a symlink or not. If it is a symlink, I dereference the relative location of the symlink and add it to the package using archiver's `.symlink(file, target)` method added in 2.0.0 (currently serverless-webpack is using 2.1.1).

## How can we verify it:
Include a large binary file in the lambda function, and then multiple symlinks pointing at that same file and then package it. The packaged zip should contain multiple copies of that binary by default, or only one copy of the binary and soft-symlinks to it if `symlinks: false` is set.

```yaml
functions:
  exampleFunction:
    handler: src/index.handler
    package:
      symlinks: false
```

## Todos:
I still need to write tests/documentation, but wanted to do those after confirming this is a PR you're willing to accept.

- [ ] Write tests
- [ ] Write documentation
- [x] Fix linting errors

### DIFF
--- a/lib/packageModules.js
+++ b/lib/packageModules.js
@@ -59,7 +59,7 @@ function zip(directory, name, followSymlinks) {
       const stats = followSymlinks ? fs.statSync(fullPath) : fs.lstatSync(fullPath);
 
       if (!stats.isDirectory(fullPath)) {
-        if (stats.isSymbolicLink()) {
+        if (followSymlinks === false && stats.isSymbolicLink()) {
           const resolvedPath = path.relative(directory, fs.realpathSync(fullPath));
           zip.symlink(filePath, resolvedPath);
         } else {
--- a/lib/packageModules.js
+++ b/lib/packageModules.js
@@ -23,7 +23,7 @@ function setArtifactPath(funcName, func, artifactPath) {
   }
 }
 
-function zip(directory, name) {
+function zip(directory, name, followSymlinks) {
   const zip = archiver.create('zip');
   // Create artifact in temp path and move it to the package path (if any) later
   const artifactFilePath = path.join(this.serverless.config.servicePath,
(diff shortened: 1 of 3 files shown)


summera:
@kevinsperrine I also needed this ability and didn't see your PR until now. It would be great to get this merged. 

I took a slightly different approach that you can see [here](https://github.com/serverless-heaven/serverless-webpack/compare/master...Loomly:add-symlink-support?expand=1). I do like the way that you provided the ability to configure at both the function and service level. One thing I found out while working on this is that `zip.append` can actually handle symlinks as you can see in my branch. It's not documented well, but it works. I think you can also use `#lstatSync` whether it's a symlink or not. Just a few ideas that might help here. Let me know if I can help in moving this forward.

j0k3r:
Could  you rebase against the master and fix conflicts? Thanks 🙏

kevinsperrine:
@j0k3r I had a little time to look at this today, and I don't think my implementation is possible anymore. The switch to bestzip abstracts the source resolving away and makes my code unable to perform the same checks for symlinks that it was able to previously.

j0k3r:
That is a sad news.
Can't we handle the symlink case in the old way?
I mean, if user define `symlink: false`, building the zip won't be performed by bestzip but by archiver.

It'll add code only for that option, but looks like some people require it.
What do you think?

PR #471, opened 2019-01-05
Adds the ability to specify an excludeFiles glob

designfrontier:
## What did you implement:

Closes #433

## How did you implement it:

Fairly trivial, just used the ignore option in glob to exclude files matching the glob.

## How can we verify it:

1. Setup a serverless project with an index.test.js and an index.js that handle a route
2. Start it up and see that you get a warning about duplicate files matching for index
3. Now pull in this PR
4. add the following to your serverless.yml file in the custom > webpack section
`excludeFiles: **/*.test.js`
5. restart serverless
6. see that the warnings are now gone!

## Todos:

- [x] Write tests
- [x] Write documentation
- [x] Fix linting errors
- [x] Make sure code coverage hasn't dropped
- [x] Provide verification config / commands / resources
- [x] Enable "Allow edits from maintainers" for this PR
- [x] Update the messages below

***Is this ready for review?:*** YES
***Is it a breaking change?:*** NO

PR #488, opened 2019-03-02
add the ability to exclude some files from the package

### LINKED ISSUE #470: Allow to skip sourcemaps from the package
# This is a Feature Proposal

## Description
Allow filtering files after webpack runs but before packaging.

When you have source maps enabled, source-maps get bundled in the deployment package (zip file) along with the compiled js bundle.

If you're using a separate Error Reporting system like [Sentry](https://sentry.io) you don't need to upload the source maps in the package zip file. This reduces the final package zip size by ~ 50%.

#### Why not disable source-maps in webpack?
If we do that, then we'll have to build once again after packaging/deploy to get the source-maps to upload to Sentry.

_**P.S. I don't mind sending a PR for this, I just want to use this issue to discuss the feature**_

## Additional Data

* ***Serverless-Webpack Version you're using***: 5.2.0
* ***Webpack version you're using***: 4.28.2
* ***Serverless Framework Version you're using***: 1.35.0
* ***Operating System***: macOS

jaydp17:
## What did you implement:

Closes #470 

It allows the user to include only a subset of files in the package.<br/>
For example, it's very handy to exclude sourcemaps from the package.

## How did you implement it:
```yaml
# serverless.yml
custom:
  webpack:
    builtAssetIncludeGlob: "**/*.{js,json}"
```

☝️this is how we use it.

We were already using [glob](https://www.npmjs.com/package/glob) package and passing `**` to include all files, this PR just makes that input to glob package as a configurable value from `serverless.yml`.

If not set in `serverless.yml` it defaults to `**`, including all files.

## How can we verify it:

1. As described above, just add the option in serverless.yml.
2. Run `sls package`
3. Extract the zip file in `.serverless` and verify that `.js.map` files are not included in the package

## Todos:

- [x] Write tests
- [x] Write documentation
- [x] Fix linting errors

### DIFF
--- a/README.md
+++ b/README.md
@@ -449,6 +449,22 @@ custom:
 This can be useful, in case you want to upload the source maps to your Error
 reporting system, or just have it available for some post processing.
 
+### Include only a subset of built assets
+
+If you don't want all built assets (i.e. js, json, js.map files, etc.) to be included
+in the package, then you can specify the glob value that'll be used to match the files
+to include in the package.
+
+```yaml
+# serverless.yml
+custom:
+  webpack:
+    builtAssetIncludeGlob: "**/*.{js,json}"
+```
+
+This can be useful when you don't want to include the source maps to make the package
+even smaller.
+
 #### Examples
 
 You can find an example setups in the [`examples`][link-examples] folder.
--- a/lib/Configuration.js
+++ b/lib/Configuration.js
@@ -14,6 +14,7 @@ const DefaultConfig = {
   packager: 'npm',
   packagerOptions: {},
   keepOutputDirectory: false,
+  builtAssetIncludeGlob: '**',
   config: null
 };
 
@@ -69,6 +70,10 @@ class Configuration {
 
   get keepOutputDirectory() {
     return this._config.keepOutputDirectory;
+  }
+
+  get builtAssetIncludeGlob() {
+    return this._config.builtAssetIncludeGlob;
   }
 
   toJSON() {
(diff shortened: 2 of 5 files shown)


designfrontier:
@jaydp17 do you mind correcting the merge conflicts outlined above? We're working our way through the backlog of pull requests. Thanks!

hassankhan:
Is this PR potentially a duplicate of #471? 

/cc @designfrontier @HyperBrain @jaydp17

HyperBrain:
Can it be that this functionality is quite dangerous? Currently webpack completely controls what is packaged, i.e. what is used is packaged. Interfering with that might render broken deployments, e.g. if you use the json-loader (which you should when using json files) and forget to include the json files.

jaydp17:
@HyperBrain this won't interfere with webpack at all, as this runs **after** webpack has finished building.
After webpack has outputted whatever assets it wanted to (i.e. bundle.js, bundle.js.map, etc.), this glob is applied to those files.

@designfrontier give me some time, I'll resolve the merge conflicts.

@hassankhan, `includeFiles` feels ambiguous, without reading the docs and just by looking at the config, you can't be sure which files does it mean.
Thus I tried to be a little descriptive with `builtAssetIncludeGlob` key, because we want to let the user know that this glob applies on the asset files outputted by webpack.

However, I'm open to suggestions, if `builtAssetIncludeGlob` doesn't feel right.

j0k3r:
We do have an option `excludeRegex` which remove some fiels from the zip.
https://github.com/serverless-heaven/serverless-webpack#exclude-files-with-regular-expression

Can this solve your problem?

PR #496, opened 2019-04-26
Improve Developer Experience

hassankhan:
## What did you implement:

Set up Git hooks to run linters/tests on a Git `pre-commit` hook.

## How did you implement it:

Added `husky`, `prettier` and `lint-staged` and relevant configuration files.

## How can we verify it:

Branch from this PR and try committing a change to any `.js` file.

## Todos:

- [ ] ~Write tests~
- [ ] ~Write documentation~
- [X] Fix linting errors


HyperBrain:
We should configure prettier exactly the same way as eslint. Your setting
seems to be a bit different

Hassan Khan <notifications@github.com> schrieb am Sa., 27. Apr. 2019, 1:31
vorm.:

hassankhan:
Thanks for looking over it, @HyperBrain 👍 I've updated everything except the `trailingComma` option because I wanted to ask if you would be open to allowing trailing commas?

prettier.config.js
```diff
@@ -0,0 +1,7 @@
+module.exports = {
+  arrowParens: 'always',
+  bracketSpacing: false,
+  semi: true,
+  singleQuote: true,
+  trailingComma: 'all',
```
HyperBrain: I'm not sure, if it is good to have two different stylers applied sequentially, because one will override the previous and might lead to failing eslint checks on PRs. If we try to apply such a mechanism, the prettier configuration should exactly mirror the eslint configuration:
```
 indent: [ error, 2, {
    MemberExpression: off
  } ]
  linebreak-style: [ error, unix ]
  semi: [ error, always ]
  quotes: [ error, single, { avoidEscape: true } ]
  no-mixed-spaces-and-tabs: error
  space-before-blocks: error
  arrow-spacing: error
  key-spacing: [ error, { afterColon: true, mode: minimum } ]
  brace-style: [ error, '1tbs' ]
  comma-spacing: [ error, { before: false, after: true } ]
  comma-style: [ error, last, { exceptions: { VariableDeclaration: true } } ]
  array-bracket-spacing: [ error, always, { singleValue: false } ]
  computed-property-spacing: [ error, never ]
  object-curly-spacing: [ error, always ]
  prefer-const: error
  no-var: error
  promise/no-nesting: off
  import/first: error
  import/newline-after-import: error
  import/no-named-as-default: off
  import/no-extraneous-dependencies: [ error, { devDependencies: true } ]
  lodash/import-scope: off
  lodash/preferred-alias: off
  lodash/prop-shorthand: off
  lodash/prefer-lodash-method: [ error, { ignoreObjects: [ BbPromise, path ] } ]
```
Currently it does not. Eslint applies much more and different restrictions even for different imported modules.
Maybe it would be enough to just run `eslint --fix` but not prettier?

hassankhan: Prettier doesn't come with a ton of options unfortunately (https://prettier.io/docs/en/options.html). That said, I wouldn't worry about ESLint/Prettier conflicting, since I've also added [`eslint-config-prettier`](https://github.com/prettier/eslint-config-prettier), which exists solely for this purpose.

package.json
```diff
@@ -61,15 +61,20 @@
...
+    "lint-staged": "^8.1.5",
     "mocha": "^5.2.0",
     "mockery": "^2.1.0",
     "nyc": "^12.0.2",
+    "prettier": "^1.17.0",
     "serverless": "^1.26.1",
     "sinon": "^6.1.3",
-    "sinon-chai": "^3.2.0"
+    "sinon-chai": "^3.2.0",
+    "webpack": "^4"
```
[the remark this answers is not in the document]
hassankhan: Removed `webpack` from `devDependencies`.

prettier.config.js
designfrontier: We use something very similar to this and it works great for us where I work. We don't run into issues with the two conflicting because of the packages mentioned above.

HyperBrain: Ok, you convinced me 😄 .

HyperBrain:
Feel free to get this in as soon as you are finished 👍

PR #502, opened 2019-05-20
default webpackConfig.node should be false

stevemao:
completely turn off the NodeStuffPlugin and NodeSourcePlugin plugins.

We don't want webpack to mock node globals

See https://codeburst.io/use-webpack-with-dirname-correctly-4cad3b265a92 and https://webpack.js.org/configuration/node

## What did you implement:

Closes #XXXXX

## Todos:

- [x] Write tests
- [ ] Write documentation
- [x] Fix linting errors

### DIFF
--- a/lib/validate.js
+++ b/lib/validate.js
@@ -141,6 +141,11 @@ module.exports = {
           filename: '[name].js',
         };
       }
+      
+      // Default node
+      if (!this.webpackConfig.node || _.isEmpty(this.webpackConfig.node)) {
+        this.webpackConfig.node = false;
+      }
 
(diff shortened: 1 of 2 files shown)


HyperBrain:
Good catch 👍 . Could you please add the check for the default value to the unit tests to keep up the coverage and complete the checklist above (TODOS)?

HyperBrain:
I'm waiting for the merge, because a fix for #505 might still go into 5.3.1 as it is urgent but this one is for 5.4.0 as it is a feature by nature, not a bug fix that does not change the expectations.

PR #520, opened 2019-06-20
Add TypeScript definition

dalen:
## What did you implement:

Add a typescript definition for the lib export.
This makes it easier to use this plugin with a webpack.config.ts
like described at https://webpack.js.org/configuration/configuration-languages/#typescript


index.d.ts
```diff
@@ -0,0 +1,10 @@
+declare module 'serverless-webpack' {
+  const lib: {
```
HyperBrain: Maybe we should list all public members of the lib object here. There are also options and the serverless object. Please see the README for details.

dalen: Good point. I've updated it to add the serverless export and the options now.

dalen:
@HyperBrain I've fixed the requested changes, could you take another look at this?

HyperBrain:
will do later this week

PR #530, opened 2019-08-07
Add support for multiple runtimes in service

brendanhonea:
## What did you implement:

Added support for projects that have lambdas running on multiple runtimes. In our case, we needed some java lambdas to run in the same project as our typescript lambdas and the plugin was trying to bundle all runtimes.

## How did you implement it:

In validate.js, added a check for runtime.startswith('node') so that the plugin only tries to package node lambdas.

## How can we verify it:

Create a serverless project that includes node lambdas, and lambdas from other supported runtimes (java, python, etc). Make sure that package.invidually is true. Master branch will try to package the other runtimes with webpack and will cause errors when invoking the lambdas. This fork will package only node lambdas with the plugin

## Todos:

- [X] Write tests
- [X] Write documentation (None needed)
- [X] Fix linting errors


ineffyble:
Any updates on this? It looks like an alternative to #256

somq:
Hi @brendanhonea, thanks for the PR.
Would be nice to also add the skipCompile flag [here](https://github.com/serverless-heaven/serverless-webpack/blob/21393845fb173a6f806b0bc2bee0be7daf0adc86/index.js#L112) in the `before:invoke:local:invoke` hook to allow sls invoke local to work with Python lambdas

```js

      'before:invoke:local:invoke': () =>
          ...
          .then(() => (this.skipCompile ? BbPromise.resolve() : this.serverless.pluginManager.spawn('webpack:compile')))
          .then(() => (this.skipCompile ? BbPromise.resolve() : this.prepareLocalInvoke)),
          // .then(this.prepareLocalInvoke),
```

Could you eventually add this to your fork?

lib/validate.js
```diff
@@ -96,8 +96,11 @@ module.exports = {
       _.merge(entries, entry);
     } else {
       _.forEach(functions, (func, index) => {
-        const entry = getEntryForFunction.call(this, functions[index], this.serverless.service.getFunction(func));
-        _.merge(entries, entry);
+        const functionDef = this.serverless.service.getFunction(func);
+        if (_.startsWith(functionDef.runtime, 'node')) {
```
miguel-a-calles-mba, line 100: What happens when the global runtime is node and the runtime is not specified per function?

miguel-a-calles-mba:
Closing. Will use this PR instead: Check for node runtimes first #579.

PR #535, opened 2019-09-02
Options to control which functions will be build with webpack and whi…

### LINKED ISSUE #255: Skip webpack on a per function basis

quassnoi:
…ch will not

closes #255

## What did you implement:

Closes #255

## How did you implement it:

I've added the configuration options to include and exclude functions that will be built using Webpack; made the `--function` option support multiple entries.

## How can we verify it:

* Create a serverless-webpack configuration with three functions
* Add one of those functions under custom / webpack / excludeFunctions, make sure Webpack does not try to build that function
* Add one of those functions under custom / webpack / includeFunctions, make sure Webpack does not try to build functions other than that
* Run `serverless webpack --function func1 --function func2`, make sure only Webpack only tries to build `func1` and `func2`

## Todos:

- [x] Write tests
- [x] Write documentation
- [x] Fix linting errors

### DIFF
--- a/README.md
+++ b/README.md
@@ -452,6 +452,41 @@ custom:
 
 This is also useful for projects that use TypeScript.
 
+#### Exclude functions from the Webpack build
+
+By default, serverless-webpack assumes that all the functions defined in the template can and will be built using Webpack.
+This is not always the case, as serverless supports multi-language stacks, where some of the functions can be implemented in other languages, like C# or Python or Go.
+To control which functions should the plugin build, you can use configuration settions:
+
+```yaml
+# serverless.yml
+custom:
+  webpack:
+    # This is useful when all your functions except some are in JS or TS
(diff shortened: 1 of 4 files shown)


quassnoi:
@HyperBrain @thenikso hey guys, mind reviewing this?

We're using this feature in our scenario, would have been nice to have the real package instead of a fork in our `package.json`

Thanks!

quassnoi:
@ceilfors @franciscocpg @HyperBrain @MartinRisk @nathanchapman @rdsedmundo hey guys, are you still supporting this project?

j0k3r:
@quassnoi Could you rebase against the master and fix conflicts? Thanks 🙏

PR #579, opened 2020-04-18
Check for node runtimes first

### LINKED ISSUE #255: Skip webpack on a per function basis

smil2k:
webpack should not try to load handlers of different runtime than node.

## What did you implement:

Closes #255 or #256, #535 

## How can we verify it:

This config will work:
```yaml
provider:
  name: aws
  runtime: nodejs12.x 

functions:
  echo:
    handler: lambda.echo
    events:
      - http:
          path: /api/echo2
          method: any
  all:
    handler: backend.lambda.LambdaFunction
    runtime: java8
    events:
      - http:
          path: /api/v1/{api+}
          method: any     
```

## Todos:

- [ ] Write tests
- [ ] Write documentation
- [x] Fix linting errors

### DIFF
--- a/lib/validate.js
+++ b/lib/validate.js
@@ -96,8 +96,12 @@ module.exports = {
       _.merge(entries, entry);
     } else {
       _.forEach(functions, (func, index) => {
-        const entry = getEntryForFunction.call(this, functions[index], this.serverless.service.getFunction(func));
-        _.merge(entries, entry);
+        const loadedFunc = this.serverless.service.getFunction(func);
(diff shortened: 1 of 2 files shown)


j0k3r:
@miguel-a-calles-mba I think tests from https://github.com/serverless-heaven/serverless-webpack/pull/530 should be ported here (and then closed) and that PR be merged instead oh the other one (because it handles the global runtime)

miguel-a-calles-mba:
Works for me @j0k3r. I updated the test.

PR #613, opened 2020-07-29
Fix behavior of file rebasing when file starts with "file:./" 

j0k3r:
Superseded #474 because we got no answer from OP.

> Relative files that are located in the same directory as the `package.json`, e.g. `"file:./foo.tgz"`, were not being rebased correctly by `rebaseFileReferences` due to a bad regular expression. The regex has
been fixed in this commit.


lib/packExternalModules.js
```diff
@@ -9,7 +9,7 @@ const isBuiltinModule = require('is-builtin-module');
 const Packagers = require('./packagers');
 
 function rebaseFileReferences(pathToPackageRoot, moduleVersion) {
-  if (/^(?:file:[^/]{2}|\.\/|\.\.\/)/.test(moduleVersion)) {
+  if (/^(file:[^/]|\.\/|\.\.\/)/.test(moduleVersion)) {
```
miguel-a-calles-mba, line 12: Should there be a test to make sure `file:./` works?

j0k3r: Good idea!

j0k3r: I tried to add some tests but there were all failed with the actual changes 🤔

PR #626, opened 2020-07-31
Update all examples

j0k3r:
## What did you implement:

I updated all examples to latest deps & version.
They should all run properly and be up to date.

I've removed some of them because there weren't accurate as of today.

I've also updated the Travis build:

- Avoid double build when PR is created on upstream (a build for PR and one for push).
- Only run coverage one build (and send coverage only once instead of twice). Also don't run coverage on Appveyor as we don't use it.

Note: I targeted master so dependabot PR will be closed automatically after merging that PR.
Note 2: lot of diff stuff but mostly due to lock files being updated.

## Todos:

- [ ] Write tests
- [ ] Write documentation
- [x] Fix linting errors


.travis.yml
```diff
@@ -2,17 +2,27 @@ os: linux
 dist: xenial
 language: node_js
 
+# avoid double Travis build when the PR is created on upstream
+if: |
+  type = pull_request OR \
+  branch = master
```
miguel-a-calles-mba, line 8: Shouldn't we run builds on `master` to make sure it build properly? Some repositories treat master as unstable, but not sure whether we are doing that for this repo.

examples/babel-multiple-statically-entries/serverless.yml
```diff
@@ -6,7 +6,7 @@ plugins:
 
 provider:
   name: aws
-  runtime: nodejs6.10
+  runtime: nodejs12.x
```
miguel-a-calles-mba, line 9: Nice. Thanks!

PR #637, opened 2020-09-03
Upgrade deps

j0k3r:
## What did you implement:

Upgrade all deps

## How did you implement it:

Mostly to fix security issue but also to keep them updated.

Also:
- fix an ESLint error: `8:16  error  Do not access Object.prototype method 'hasOwnProperty' from target object  no-prototype-builtins`
- run prettier (following update)
- update `format` script with `$(pwd)` to fix issue with higher version of ESLint (>5) see https://github.com/prettier/prettier-eslint-cli/issues/205#issuecomment-557746077
- prettier now run on every js files

## Todos:

- [ ] Write tests
- [ ] Write documentation
- [x] Fix linting errors

### DIFF
--- a/examples/babel-multiple-statically-entries/first.js
+++ b/examples/babel-multiple-statically-entries/first.js
@@ -2,10 +2,10 @@ export const hello = (event, context, cb) => {
   const p = new Promise((resolve, reject) => {
     resolve('success');
   });
-  p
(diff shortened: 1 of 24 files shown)


miguel-a-calles-mba (approved):
Looks good to me.

PR #680, opened 2021-01-06
Feat/concurrency

vicary:
## What did you implement:

Supersedes #517, adding concurrency support for more than one thread.

## How did you implement it:

Please see https://github.com/serverless-heaven/serverless-webpack/pull/517#issuecomment-653547396, basically I did two changes:

1. Replacing `serizliedCompile: Boolean` with `concurrency: Number`
2. Mapping `serializedCompile: true` to `concurrency: 1` for backward compatibility

## How can we verify it:

Use the config below when packaging:

```
package:
  individually: true

custom:
  webpack:
    concurrency: 5
```

## Todos:

- [x] Write tests
- [x] Write documentation
- [x] Fix linting errors
- [x] Make sure code coverage hasn't dropped
- [x] Provide verification config / commands / resources
- [x] Enable "Allow edits from maintainers" for this PR
- [x] Update the messages below

***Is this ready for review?:*** Yes
***Is it a breaking change?:*** No

PR #517, opened 2019-06-14
Serialized compile to address #299

asprouse:
## What did you implement:

Add the option to serialize webpack compilation when packaging individually. This addresses concerns in #299 and improves compilation speed significantly. In our project with 25 function this decreased the build time from ~400s to ~265s when run with the following command:
```
time node --max-old-space-size=4096 node_modules/.bin/sls package --verbose
```
While using `thread-loader` + `cache-loader` might be able to optimize accross the multiple builds it takes significant effort to get to a performant build. Running the webpack compilation in series appears to have better performance out of the box, so much so that I suggest it be the default behavior of this plugin. This PR maintains the existing multi-compile as the default and uses `custom.webpack.serializedCompile` to enable the serial build.

## How did you implement it:
Modified `compile.js` to use `BbPromise.mapSeries` instead of webpack's multi-compiler functionality. This cut down

## How can we verify it:
Set `custom.webpack.serializedCompile` to `true` in `serverless.yml` and run the following command:
```
time node --max-old-space-size=4096 node_modules/.bin/sls package --verbose
```
Now set the `custom.webpack.serializedCompile` to `false` or omit the option entirely to revert to the default behavior and run the same command. Verify that both compile successfully and observe a significant speed improvement when a project with many functions is serialized.

## Todos:

- [x] Write tests
- [x] Write documentation
- [x] Fix linting errors
- [x] Make sure code coverage hasn't dropped
- [x] Provide verification config / commands / resources
- [x] Enable "Allow edits from maintainers" for this PR
- [x] Update the messages below

***Is this ready for review?:*** YES
***Is it a breaking change?:*** NO

PR #681, opened 2021-01-13
Add concurrency support for more than one thread

vicary:
## What did you implement:

Adding concurrency support for more than one thread.

Conditionally supersedes #680 if `release/5.4.0` is more preferable than `master`.

## How did you implement it:

Please see https://github.com/serverless-heaven/serverless-webpack/pull/517#issuecomment-653547396, basically I did two changes:

1. Replacing `serizliedCompile: Boolean` with `concurrency: Number`
2. Translates `serializedCompile: true` to `concurrency: 1` for backward compatibility

## How can we verify it:

Use the config below when packaging:

```
package:
  individually: true

custom:
  webpack:
    concurrency: 5
```

## Todos:

- [x] Write tests
- [x] Write documentation
- [x] Fix linting errors

***Is it a breaking change?:*** No


coyoteecd:
@miguel-a-calles-mba @j0k3r can you have a look and merge this? I am using vicary's fork for half a year and would love to switch back to the official package.

j0k3r:
@coyoteecd we are in the process of cleaning up stuff to release 5.4.0 and this will be part of it

j0k3r:
Sorry I made a mistake while merging `release/5.4.0` into master and removed it. The removal automatically closed that PR. It should now be rebased against the master and fix conflicts.

vicary:
Not sure if I'm doing it wrong, but I found myself learning the backstory every conflict or two in the process of rebasing. This takes a day or two, could only do this on weekends.

vicary:
@j0k3r During my rebase I found that the current master has these patterns.

```js
// tests/packageModules.test.js
        return BbPromise.each([ '1.18.1', '2.17.0', '10.15.3' ], version => {
          getVersionStub.returns(version);
          return expect(module.packageModules()).to.be.fulfilled.then(() => BbPromise.all([]));
        }).then(() =>
          BbPromise.each([ '1.17.0', '1.16.0-alpha', '1.15.3' ], version => {
            getVersionStub.returns(version);
            return expect(module.packageModules()).to.be.fulfilled.then(() => BbPromise.all([]));
          })
        );
```

It has a strong code-smell here `.then(() => BbPromise.all([]));` which should either not there, or some tests are missed during rebase.

Not meant to make things more complicated than it should, please feel free to ignore this if it's intended.

I'll be rebasing to master as-is, just a friendly heads-up.

j0k3r:
@vicary I've no idea but it seems useless 🤔 Have you tried to remove the useless part `.then(() => BbPromise.all([]));` to see if tests are still green?

vicary:
> @vicary I've no idea but it seems useless 🤔 Have you tried to remove the useless part `.then(() => BbPromise.all([]));` to see if tests are still green?

They must produce the same result, because it does not add any logics besides an empty promise resolution.

From the rebase conflicts I can see there was other assertions inside the arrays. I am not sure if these sub-assertions are meant to be removed, if so, the code segment can be safely removed.

vicary:
The rebase is finished, this should be ready for merging.

lib/Configuration.js
```diff
@@ -15,7 +15,7 @@ const DefaultConfig = {
   packagerOptions: {},
   keepOutputDirectory: false,
   config: null,
-  serializedCompile: false
+  concurrency: Infinity
```
j0k3r, line 18: So the default value is infinity? Shouldn't this be _to the number of available cores_?

j0k3r:
Awesome job for the rebase @vicary 👏

lib/Configuration.js
vicary: You're correct.

lib/validate.js
```diff
@@ -190,10 +190,11 @@ module.exports = {
       // In case of individual packaging we have to create a separate config for each function
       if (_.has(this.serverless, 'service.package') && this.serverless.service.package.individually) {
         this.multiCompile = true;
-        this.serializedCompile = this.configuration.serializedCompile;
         this.options.verbose &&
           this.serverless.cli.log(
-            `Using ${this.serializedCompile ? 'serialized' : 'multi'}-compile (individual packaging)`
+            `Using ${
+              this.configuration.concurrency !== Infinity ? 'concurrent' : 'multi'
```
j0k3r, line 196: Shouldn't this be updated as well? Since concurrency default value is no more `Infinity`

j0k3r (commented):
Also, I think some tests should be updated as well about `{ concurrency: Infinity }`. Don't you think?

lib/validate.js
vicary: Before `serializedCompile` is available, `serverless-webpack` blasts everything at once in the same process, that's where the `Infinity` comes from.

vicary:
> Also, I think some tests should be updated as well about `{ concurrency: Infinity }`. Don't you think?

I agree that using `Infinity` may causes confusions. In fact `compile.test.js` is largely testing if the compiler has even run once, `concurrency` config there is mostly for cosmetic reasons.

To reduce ambiguity, how about we change the tests to use `1` instead?

lib/validate.js
vicary: I just had a deeper look on this. When `package.individually === true` it set `multiCompile = true`, my searches tells me this property is no longer used anywhere else. In fact the original `Infinity` behavior must be explicitly forced via a `${file:...}` reference in `serverless.yml`.

Do you think it's a good idea to remove all instance of *multiCompile* and change the log to the following format instead? 
```javascript
`Individually packaging with concurrency at ${this.configuration.concurrency} entries a time.`
```

j0k3r: Looks like `multiCompile` was removed here https://github.com/serverless-heaven/serverless-webpack/pull/517
So yes, good idea to change the log 👍

vicary: All done

j0k3r (approved):
Good job!

PR #686, opened 2021-02-01
Support noFrozenLockfile options

tgroutars:
## What did you implement:

We have a monorepo with several serverless packages. Given the way we do CI and the way `lerna` deals with lockfiles, we actually don't need and don't have an up to date `yarn.lock`, but for our usecase it doesn't really matter.

This adds the option `noFrozenLockfile` to `packagerOptions`, allowing to run `yarn install` without the `--no-frozen-lockfile` option. This is as suggested in this comment https://github.com/serverless-heaven/serverless-webpack/issues/369#issuecomment-381519788 

I also fixed the documentation for `ignoreScripts` in the README, which mentioned a default value of `true` when it is in fact `false`

## How did you implement it:

Added the option in the `install` function in `yarn.js`

## How can we verify it:

I added tests to make sure the option was omitted when using `noFrozenLockfile`

## Todos:

- [x] Write tests
- [x] Write documentation
- [ ] Fix linting errors

### DIFF
--- a/README.md
+++ b/README.md
@@ -361,9 +361,10 @@ Using yarn will switch the whole packaging pipeline to use yarn, so does it use 
 
 The yarn packager supports the following `packagerOptions`:
 
-| Option        | Type | Default | Description |
-|---------------|------|---------|-------------|
-| ignoreScripts | bool | true    | Do not execute package.json hook scripts on install |
+| Option           | Type | Default  | Description                                         |
+|------------------|------|----------|-----------------------------------------------------|
+| ignoreScripts    | bool | false    | Do not execute package.json hook scripts on install |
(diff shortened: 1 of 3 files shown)


tgroutars:
@j0k3r Still nothing in the sidebar to allow edit from maintainers 🤔 
Is that maybe a beta feature?
Screenshot 2021-02-01 at 08 50 24

j0k3r:
I think the problem is that your fork is made from an organization and that feature isn't supported.
https://github.com/isaacs/github/issues/1681

Any chance to create the PR from your own account? Or just fix the small conflicts and it'll be fine :)

tgroutars:
Ah makes sense, I'll open it from my personal account. Don't want to mess up anything 😅

PR #655, opened 2020-11-04
Support noFrozenLockfile option

tgroutars:
## What did you implement:

We have a monorepo with several serverless packages. Given the way we do CI and the way `lerna` deals with lockfiles, we actually don't need and don't have an up to date `yarn.lock`, but for our usecase it doesn't really matter.

This adds the option `noFrozenLockfile` to `packagerOptions`, allowing to run `yarn install` without the `--no-frozen-lockfile` option. This is as suggested in this comment https://github.com/serverless-heaven/serverless-webpack/issues/369#issuecomment-381519788 

I also fixed the documentation for `ignoreScripts` in the README, which mentioned a default value of `true` when it is in fact `false`

## How did you implement it:

Added the option in the `install` function in `yarn.js`

## How can we verify it:

I added tests to make sure the option was omitted when using `noFrozenLockfile`

## Todos:

- [x] Write tests
- [x] Write documentation
- [ ] Fix linting errors
- [ ] Make sure code coverage hasn't dropped
- [ ] Provide verification config / commands / resources
- [ ] Enable "Allow edits from maintainers" for this PR
- [ ] Update the messages below

***Is this ready for review?:*** YES
***Is it a breaking change?:*** NO

PR #687, opened 2021-02-01
Support noFrozenLockfile options

tgroutars:
## What did you implement:

We have a monorepo with several serverless packages. Given the way we do CI and the way `lerna` deals with lockfiles, we actually don't need and don't have an up to date `yarn.lock`, but for our usecase it doesn't really matter.

This adds the option `noFrozenLockfile` to `packagerOptions`, allowing to run `yarn install` without the `--no-frozen-lockfile` option. This is as suggested in this comment https://github.com/serverless-heaven/serverless-webpack/issues/369#issuecomment-381519788 

I also fixed the documentation for `ignoreScripts` in the README, which mentioned a default value of `true` when it is in fact `false`

## How did you implement it:

Added the option in the `install` function in `yarn.js`

## How can we verify it:

I added tests to make sure the option was omitted when using `noFrozenLockfile`

## Todos:

- [x] Write tests
- [x] Write documentation
- [ ] Fix linting errors

### DIFF
--- a/README.md
+++ b/README.md
@@ -361,9 +361,10 @@ Using yarn will switch the whole packaging pipeline to use yarn, so does it use 
 
 The yarn packager supports the following `packagerOptions`:
 
-| Option        | Type | Default | Description |
-|---------------|------|---------|-------------|
-| ignoreScripts | bool | true    | Do not execute package.json hook scripts on install |
+| Option           | Type | Default  | Description                                         |
+|------------------|------|----------|-----------------------------------------------------|
+| ignoreScripts    | bool | false    | Do not execute package.json hook scripts on install |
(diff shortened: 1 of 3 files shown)


tgroutars:
@j0k3r

j0k3r:
Following #655 & #686

j0k3r (approved):

j0k3r:
Thanks @tgroutars!

tgroutars:
No, thank _you_ 🙏 
Do you already know in which release this will be included?

j0k3r:
I added it to the next release the 5.4.0.
There are still some PRs we would like to push in that release, but it's coming!

PR #688, opened 2021-02-04
Testing with Node.js 14.x

j0k3r:
As [AWS Lambda now supports Node.js 14.x](https://aws.amazon.com/fr/blogs/compute/node-js-14-x-runtime-now-available-in-aws-lambda/), just checking the library is ok with it.

Also cleanup GitHub Actions files


miguel-a-calles-mba (approved):
Looks good.

Issue #431, opened 2018-08-02
Using yarn workspaces leads to wrong behavior of some subsystems


otbe:
# This is a (Bug Report / Feature Proposal) (maybe both?)
## Description
I use `serverless-webpack` in a monorepo with several services. My structure is something like this:

```
├── packages/
│   ├── service1
│   └── service2
├── package.json
├── yarn.lock
``` 

Each service has is own `package.json`, `serverless.yml` and `webpack.config.js`.
This works so far but there are 2 issues with this setup:

1. I get a lot of `WARNING: Could not check for peer dependencies of <dep>` warnings. I guess this is related to [this](https://github.com/serverless-heaven/serverless-webpack/blob/d542eb63bad77ccf42edea7f7a3abcdbb14f6720/lib/packExternalModules.js#L83) which just looks at the wrong place for `node_modules`. In `webpack-node-externals` I can define a folder it will look into, but that feels wrong too. Maybe we need a node-like resolution algo here?

2. If I'm using the `yarn` packager ,`serverless-webpack` will not find my lockfile because its some folders upwards in the tree.

I'm using the latest serverless/serverless-webpack version.

I really wonder if I should use `serverless-webpack`in this way and If so how?

Thank you for all your great work in this field 👍

aneilbaboo:
I have a similar monorepo structure, where I have multiple subfolders each with their own serverless.yml and webpack configs. 

I'm seeing a strange issue on CircleCI. When I first run a webpack build, it seems to ignore directives in the webpack configuration file, and instead attempts to transpile files where were explicitly excluded, including files from other subdirectories. The excluded files contain issues which cause webpack to faill. 

The weird thing is that running the command again (i.e., `yarn sls webpack`) succeeds (i.e., on the second run, the excluded files don't trigger errors). Again, this only happens in my CI environment. But both serverless, webpack and serverless-webpack are the same in both my local and CI environments.

I thought perhaps some state was being generated in either `.serverless/` or `.webpack/` which caused webpack to ignore the problem files on the second run, but deleting these doesn't cause the problem to reappear. 

I'm not sure if this is related to the issue you're seeing.

esetnik:
I have the same issue using `serverless-webpack` with yarn workspaces. In my case it appears that only a warning is generated `WARNING: Could not check for peer dependencies of <dep>`

sthuber90:
I observe the same output, and `yarn.lock`file is not found in the project root, which for me is a not wanted behavior

[...]

PR #689, opened 2021-02-09
Add ability to specify the `node_modules` relative dir

ypxing:
## What did you implement:

serverless-webpack uses `node_modules` to find out all `peerDependencies` but the assumption is that `node_modules` is in the same directory as package.json. But this is not always true specially for monorepo case where node_modules are usually in one parent directory.

This PR is to remove this assumption by adding one customer configuration `nodeModulesRelativeDir`. By default it follows current behaviour.

It can addresses issue #431 

## How can we verify it:
Testing cases has been updated to cover with/without the new configuration.
README has been updated.

Testing shows below error message is gone and peerDependencies can be packed properly

```
WARNING: Could not check for peer dependencies
```

## Todos:

- [x] Write tests
- [x] Write documentation
- [x] Fix linting errors

### DIFF
--- a/lib/packExternalModules.js
+++ b/lib/packExternalModules.js
@@ -66,7 +66,7 @@ function removeExcludedModules(modules, packageForceExcludes, log) {
  * Resolve the needed versions of production dependencies for external modules.
  * @this - The active plugin instance
  */
-function getProdModules(externalModules, packagePath, dependencyGraph, forceExcludes) {
+function getProdModules(externalModules, packagePath, nodeModulesRelativeDir, dependencyGraph, forceExcludes) {
   const packageJsonPath = path.join(process.cwd(), packagePath);
   const packageJson = require(packageJsonPath);
   const prodModules = [];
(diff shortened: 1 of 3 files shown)


lib/packExternalModules.js
```diff
@@ -83,14 +83,14 @@ function getProdModules(externalModules, packagePath, dependencyGraph, forceExcl
     if (moduleVersion) {
       prodModules.push(`${module.external}@${moduleVersion}`);
 
+      let nodeModulesBase = path.dirname(path.join(process.cwd(), packagePath));
+
+      if (nodeModulesRelativeDir) nodeModulesBase = path.join(process.cwd(), nodeModulesRelativeDir);
```
j0k3r, line 88: I prefer that kind of way for readability
```suggestion
      if (nodeModulesRelativeDir) {
        nodeModulesBase = path.join(process.cwd(), nodeModulesRelativeDir);
      }
```

j0k3r (commented):
Can't we have an error or a warning when `nodeModulesRelativeDir` is not a real directory? Like for a misconfiguration?

lib/packExternalModules.js
ypxing: Thanks @j0k3r 
comments have been addressed.

j0k3r:
Can you rebase against the master and fix conflicts? Thanks 🙏

ypxing:
Sorry for late response. Rebase is done

j0k3r (approved):

PR #691, opened 2021-02-15
Update CHANGELOG for 5.4.0

j0k3r:
It's been too long since the latest release and some PRs are too old to be part of the 5.4.0 release. So let's release it.
PR with the 5.4.0 milestone will be shipped in 5.5.0 or maybe 5.4.1, we'll later.

Here is a proposal for the text for the release:

```
New options:
- `noFrozenLockfile` (boolean) under the `packagerOptions` for the Yarn packager allows you to not have an up to date `yarn.lock`. See #687
- `excludeRegex` (string) allows you to filter files that match the regex before adding to the zip. See #612
- `serializedCompile` (boolean) allows you to run each webpack buid one at a time to reduce memory usage. See #517
- `allowCustomRuntime` (boolean) added to the function level in your `serverless.yml` allows you force the packaging when you are using a nodejs custom runtime ONLY. See #675
- `concurrency` (integer) allow to set the concurrency (defaults to the number of available cores), only works when packing functions individually. See #681 

New arguments:
- the `--no-build` used when deploying won't re-build everything if you already build it (using `sls webpack` for example). See #560

Optimizations:
- switched to `bestzip` instead of `archiver` to use natine zip command if it exists. See #596
- speed up cleanup process when removing `.webpack` folder. See #462
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,11 +1,18 @@
 # Change Log
 
 * 5.4.0
-  * Skip compile & packaging if --no-build is set [#560](https://github.com/serverless-heaven/serverless-webpack/pull/560)
+  * Skip compile & packaging if `--no-build` is set [#560](https://github.com/serverless-heaven/serverless-webpack/pull/560)
   * Serialized compile to address [#299](https://github.com/serverless-heaven/serverless-webpack/pull/299) [#517](https://github.com/serverless-heaven/serverless-webpack/pull/517)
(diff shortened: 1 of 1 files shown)


thomaschaaf:
Is there any way we can support you on getting this release shipped? The changelog looks good :)

miguel-a-calles-mba (approved):

j0k3r:
Sorry @miguel-a-calles-mba I just merged one last PR for that version. Can you review again? 🙏

miguel-a-calles-mba (approved):

PR #735, opened 2021-03-18
Display the zip method used

j0k3r:
Since 5.4.0 we are using `bestzip` which can improve performance by using native `zip` method if available instead of the node one.
But we don't know which method is used during the build, so improving a small log to know that information.

```
Serverless: Zip service (using native zip method): /project/.webpack/service [11882 ms]
```

Also:
- run `npm audit fix` to fix some vulnerabilities.
- improve / cleanup return data from mocks

### DIFF
--- a/lib/packageModules.js
+++ b/lib/packageModules.js
@@ -3,7 +3,7 @@
 const BbPromise = require('bluebird');
 const _ = require('lodash');
 const path = require('path');
-const bestzip = require('bestzip');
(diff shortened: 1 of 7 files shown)


miguel-a-calles-mba (approved):
Not sure why the coverage dropped. I don't see any reduced coverage in the changes.

j0k3r:
I think the drop is caused by the ternary condition because the `hasNativeZip` mock only returns `false`

PR #746, opened 2021-03-26
Fix external modules with webpack 5

### LINKED ISSUE #651: External modules are not packaged with webpack 5
# This is a Bug Report

## Description

Since updating to the release version of webpack 5, it seems like external modules are no longer packaged. I investigated a bit and in `packExternalModules#getExternalModules` I don't see the externals I specify in my webpack config. I checked webpack 5 release notes but could not find a not about that and how to fix it. As a workaround I was able to use `stats.compilation.modules` instead of `chunks` and it does include the externals. Not sure if this is a proper fix though.

N/A

## Additional Data

* ***Serverless-Webpack Version you're using***: 5.3.5
* ***Webpack version you're using***: 5.1.3
* ***Serverless Framework Version you're using***: 2.8.0
* ***Operating System***: MacOS 10.15.7
* ***Stack Trace (if available)***: N/A

janicduplessis:
## What did you implement:

Closes #651

Seems like the issue happens because external modules are now concatenable (see https://github.com/webpack/webpack/releases/tag/v5.0.0-beta.31). What I found is using compilation.modules instead of chunks.modulesIterable includes all modules before concatenation so the issue does not happen. This also removed usages of deprecated module.issuer api.

## How did you implement it:

- Use compilation.modules instead of chunks.moduleIterable
- Use ModuleGraph api instead of deprecated module.issuer

## How can we verify it:

Using webpack 5, package an app with some external packages, make sure these packages are included in the function zip.

```js
// webpack.config.js
{
  ...
  externals: ['sharp', 'mjml']
}
```

Note: This wasn't tested with webpack 4 so if anyone has a webpack 4 setup and wants to test that would be awesome.

## Todos:

- [x] Write tests
- [ ] Write documentation
- [x] Fix linting errors

### DIFF
--- a/lib/packExternalModules.js
+++ b/lib/packExternalModules.js
@@ -130,7 +130,7 @@ function getProdModules(externalModules, packagePath, dependencyGraph, forceExcl
         const originInfo = _.get(dependencyGraph, 'dependencies', {})[module.origin] || {};
         moduleVersion = _.get(_.get(originInfo, 'dependencies', {})[module.external], 'version');
         if (!moduleVersion) {
-          // eslint-disable-next-line lodash/path-style 
+          // eslint-disable-next-line lodash/path-style
           moduleVersion = _.get(dependencyGraph, [ 'dependencies', module.external, 'version' ]);
         }
         if (!moduleVersion) {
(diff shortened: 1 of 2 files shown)


thenickdude:
Thanks for this, this fixes my serverless-webpack project!

lib/packExternalModules.js
```diff
@@ -182,36 +182,42 @@ function isExternalModule(module) {
   return _.startsWith(module.identifier(), 'external ') && !isBuiltinModule(getExternalModuleName(module));
 }
 
+/**
+ * Gets the module issuer. The ModuleGraph api does not exists in webpack@4
+ * so falls back to using module.issuer.
+ */
+function getIssuerCompat(moduleGraph, module) {
+  if (moduleGraph != null) {
```
j0k3r, line 190: It should be enough, don't you think?

```suggestion
  if (moduleGraph) {
```

lib/packExternalModules.js
```diff
@@ -182,36 +182,42 @@ function isExternalModule(module) {
...
+    return moduleGraph.getIssuer(module);
+  } else {
+    return module.issuer;
+  }
```
j0k3r, line 194: You don't need the `else` here as the previous command already exited the function

```suggestion
  }

  return module.issuer;
```

lib/packExternalModules.js
```diff
@@ -182,36 +182,42 @@ function isExternalModule(module) {
...
-
-    // Explore each module within the chunk (built inputs):
-    for (const module of chunk.modulesIterable) {
-      if (isExternalModule(module)) {
-        externals.add({
-          origin: _.get(findExternalOrigin(module.issuer), 'rawRequest'),
-          external: getExternalModuleName(module)
-        });
-      }
+  for (const module of compilation.modules) {
```
j0k3r, line 212: Just wondering if this is compatible with webpack 4? (or if it's related to what you said about not testing it on webpack 4 already?)

janicduplessis: It should be https://github.com/webpack/webpack/blob/webpack-4/lib/Compilation.js#L494. But yeah ideally someone would test on webpack 4 just to make sure.

j0k3r: I just tested and it worked fine on webpack4

j0k3r:
Looks great, can you squash your commits?

janicduplessis:
@j0k3r done!

j0k3r (approved):

j0k3r:
Good job @janicduplessis!

Issue #404, opened 2018-05-30
Cannot define webpack configuration as typescript


chanlito:
I created a `webpack.config.ts` file, but the plugin seems to ignore it.

HyperBrain:
Can you please use the issue template that is shown when opeining an issue and put in all information that can help to reproduce and analyze the issue? Examples are: your config.ts file, the used webpack and plugin versions, etc.

However, specifying a ts config should be possible because the plugin automatically should use `ts-node` in this case. With more informatino I could try to find it.

chanlito:
BTW @HyperBrain also consider updating current ts-node version is way behind. I'll try to create a repo.

HyperBrain:
Thanks for the info. As I do not use TS in our projects, I did not recognize that. Do you know if there is anything that has to be checked when updating ts-node? I saw that it is a double major version jump.

chanlito:
Yes, ts-node v6 has typescript-checking by default, which for some people is good and some is bad, like me I just want transpileOnly for faster builds because vscode editor show typescript errors pretty well already.

chanlito:
I create a repo https://github.com/chanlito/serverless-typescript-template you can clone it and try.

martinjlowm:
It works fine here (apart from a syntax error in `tsconfig.json` and the following)... with `serverless.yml`'s `webpackConfig` pointing to `webpack.config.ts`.

`serverless-webpack` does not appear to support `export default`-config at this time. See https://github.com/serverless-heaven/serverless-webpack/blob/016415926b0d0814fa94638e390f59eca3713d61/lib/validate.js#L114

It results in
```
{
  default: {
    ...config
  }
}
```

rather than
```
{
  ...config
}
```

You may use `module.exports = config` instead.

HyperBrain:
@chanlito Can this be closed?

PR #766, opened 2021-04-06
Prevent ts-node being registered twice

apancutt:
## What did you implement:

Relates to #404

It's currently not possible to use a `webpack.config.ts` file if also using a `serverless.ts` file. This is because `ts-node/register` is called twice resulting in `ts-node` trying to compile compiled Typescript.

Specifically, these errors are emitted:

```
api_1    |   TSError: ⨯ Unable to compile TypeScript:
api_1    |   webpack.config.ts(2,5): error TS7022: '__importDefault' implicitly has type 'any' because it does not have a type annotation and is referenced directly or indirectly in its own initializer.
api_1    |   webpack.config.ts(2,67): error TS7006: Parameter 'mod' implicitly has an 'any' type.
api_1    |   webpack.config.ts(45,29): error TS7006: Parameter 'resource' implicitly has an 'any' type.
```

## How did you implement it:

Using the suggestion posted [here](https://github.com/TypeStrong/ts-node/issues/846#issuecomment-631828160).

## How can we verify it:

1. Create a `serverless.ts` file AND a `webpack.config.ts` file
2. See errors above when trying to run any command that invokes serverless-webpack
3. Apply patch in this PR
4. Errors should disappear

## Todos:

- [x] Write tests
- [x] ~Write documentation~ (n/a?)
- [x] Fix linting errors
- [x] ~Provide verification config / commands / resources~ (n/a?)

### DIFF
--- a/index.js
+++ b/index.js
@@ -26,11 +26,13 @@ class ServerlessWebpack {
     this.options = options;
 
     if (
-      (_.has(this.serverless, 'service.custom.webpack') &&
-        _.isString(this.serverless.service.custom.webpack) &&
-        _.endsWith(this.serverless.service.custom.webpack, '.ts')) ||
-      (_.has(this.serverless, 'service.custom.webpack.webpackConfig') &&
-        _.endsWith(this.serverless.service.custom.webpack.webpackConfig, '.ts'))
+      (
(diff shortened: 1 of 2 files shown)


j0k3r (approved):
Looks ok to me, thanks for the PR!

PR #772, opened 2021-04-07
Fix tests failing with Serverless 2.x

nponeccop:
Closes #771

## What did you implement:

Closes #771

## How did you implement it:

The "stub" errors were essentially an attempt to stub a non-existent property. Apparently there is no property and it isn't necessary to stub it. One remaining failing test was a trivial `null` vs `undefined` rot.

## How can we verify it:

By code review. Note that I did it without knowing what I'm doing.

## Todos:

- [ ] Write tests
- [ ] Write documentation
- [ ] Fix linting errors
- [ ] Make sure code coverage hasn't dropped
- [ ] Provide verification config / commands / resources
- [ ] Enable "Allow edits from maintainers" for this PR
- [ ] Update the messages below

***Is this ready for review?:*** YES
***Is it a breaking change?:*** YES

It is a breaking change that serverless 2.x is now required through `package.json`. I'm not sure if anyone still uses Serverless 1.x though. There are also PRs to make `serverless-webpack`  less broken with 2.x

PR #775, opened 2021-04-08
Add Serverless v2 compatibility 

### LINKED ISSUE #652: Question: does this support Serverless framework v2?
The very first line of the README states: "A Serverless v1.x plugin..."

I've upgraded to Serverless v2 and things seem to work. 

I'm just checking, does the README need to be updated or is there an issue? Couldn't find anything in the issues list but it's hard set of keywords to search for.

### LINKED ISSUE #771: Tests are against an ancient 1.x version of Serverless
# This is a Bug Report

## Description

1. Bump the `serverless` version in `package.json`
2. `npm ci`
3. `npm test`

## What went wrong?

Tests were red. With `Cannot stub non-existent property servicePath` error message

## What did you expect should have happened?

Green tests

## What was the config you used?

N/A

##  What stacktrace or error message from your provider did you see?

```
 220 passing
9 failing
```

## Additional Data

- ***Serverless-Webpack Version you're using***:  Current master.
- ***Operating System***: Linux 
- ***Stack Trace (if available)***: Available but useless.

j0k3r:
Fix #771
Fix #652
Supersed #772 (thanks @nponeccop I've cherry-picked your commit 🤝)

Add a dedicated test to run the test suite against Serverless v1 (the only part related to that is in `tests/validate.test.js`), just in case.

Also:
- use `ubuntu-20-04` instead of `ubuntu-latest`
- update `actions/setup-node@v1` to v2
- remove duplicate tests suite
- run `npm run format`

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -9,7 +9,7 @@ on:
 
 jobs:
   eslint:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20-04
     name: ESLint
 
     steps:
@@ -19,7 +19,7 @@ jobs:
           fetch-depth: 2
 
       - name: "Install Node.js"
-        uses: actions/setup-node@v1
+        uses: actions/setup-node@v2
         with:
           node-version: "12"
 
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -9,7 +9,7 @@ on:
 
 jobs:
   eslint:
-    runs-on: ubuntu-20-04
+    runs-on: ubuntu-20.04
     name: ESLint
 
     steps:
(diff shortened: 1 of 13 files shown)


miguel-a-calles-mba (approved):

nponeccop:
> thanks @nponeccop I've cherry-picked your commit

Yep, it wasn't actually meant to be used as is, but merely to demonstrate that 2.x compatibility is a low-hanging fruit.

Another thought is that can we use something like "^1.xx || ^2.33" as the constraint? You fixed the suite in a way so 1.x is still supported, so can we have a relaxed constraint in `package.json`?

j0k3r:
@nponeccop I though about it too, but as Serverless is a `devDependencies` we don't really care to put a range in the version. We only need to ensure the plugin is compatible.

Issue #744, opened 2021-03-25
Directory names that start with a dot are not being included in the zip file


onhate:
# Bug

## Description

After upgrading from 5.3.5 to 5.4.0 I started getting the error below, I noticed that archived was replaced by bestzip and this is exactly what is failing, not sure what is causing this issue but I use hidden folders (starting with dot) for auto generated code that is packaged by serverless and bestzip may not be able to find the folder.

* What stacktrace or error message from your provider did you see?

```
zip error: Nothing to do! (../onAppInstanceSyncStorage.zip)
 
  Error --------------------------------------------------
 
  Error: Unexpected exit code from native zip: 12
   executed command 'zip --quiet --recurse-paths ../onAppInstanceSyncStorage.zip'
   executed in directory '/home/runner/work/hallway/hallway/server/server/dist/onAppInstanceSyncStorage'
      at ChildProcess.<anonymous> (/home/runner/work/hallway/hallway/node_modules/bestzip/lib/bestzip.js:86:13)
      at ChildProcess.emit (events.js:314:20)
      at ChildProcess.EventEmitter.emit (domain.js:483:12)
      at maybeClose (internal/child_process.js:1022:16)
      at Process.ChildProcess._handle.onexit (internal/child_process.js:287:5)
```

## Additional Data

* ***Serverless-Webpack Version you're using***: 5.4.0
* ***Webpack version you're using***: 4.44.2
* ***Serverless Framework Version you're using***: 1.83.3
* ***Operating System***: Any

j0k3r:
I'm sorry about that error.
Is there any chance you can provide a simple project to reproduce the error? So we can track it down and update serverless-webpack or bestzip to fix the issue.
Thanks

onhate:
let me try, my setup is pretty complex

ageoff:
I was also seeing this when version bumped to 5.4.0. Same setup with .<folder> names

j0k3r:
Same question then: Is there any chance you can provide a simple project to reproduce the error?

christiantakle:
@j0k3r Here you go

https://github.com/christiantakle/serverless-webpack-bug

j0k3r:
@christiantakle thanks for the repo, it helped me a lot.
In fact, when we zip files, we use `*` to match all files. But, [according to the doc](https://github.com/nfriedly/node-bestzip#dotfiles), it does not include dotfiles by default.

[...]

PR #777, opened 2021-04-08
fix excludeRegex by actually excluding files

### LINKED ISSUE #776: excludeRegex not excluding files
# This is a (Bug Report / ~~Feature Proposal~~)

## Description

For bug reports:
* What went wrong?
When running `sls package --verbose`,  `excludeRegex` option [added](612) in v5.4.0 seems to be ignoring the regex and still adds all files to the output zip files.

Files that match `excludeRegex` should be excluded from the packaged zip files.

```typescript
webpack: {
            keepOutputDirectory: true,
            excludeRegex: /\.ts|test|\.map/
        }
```
(I'm using `serverless.ts` instead of `serverless.yml` but that shouldn't matter I believe)

`sls package --verbose` does not throw any explicit errors.

For feature proposals:
* What is the use case that should be solved. The more detail you describe this in the easier it is to understand for us.
* If there is additional config how would it look

* [Initial implementation](612)

## Additional Data

* ***Serverless-Webpack Version you're using***: 5.4.1
* ***Webpack version you're using***: 5.6.0
* ***Serverless Framework Version you're using***: 2.10.0
* ***Operating System***: Windows 10

l1b3r:
## What did you implement:

Closes #776 

## How did you implement it:

* When querying a directory for files using `glob`, add a `nodir: true` parameter to have only actual files in the resulting `files` array.
* Filter the files using `excludeRegex` option if specified. If verbose, log the amount of excluded files.
* Then, when zipping files with `bestzip`, pass `files` instead of `"*"` so that only specified files end up in the zip.

## How can we verify it:

webpack.config.js
```javascript
module.exports = {
    mode: 'production',
    target: 'node',
    devtool: 'source-map',
    resolve: {
        extensions: ['.mjs', '.ts', '.js']
    },
    output: {
        libraryTarget: 'commonjs2',
        filename: '[name].js'
    },
    externals: /^aws-sdk(\/.*)?/,
    // Place all node_modules in a separate vendor chunk
    optimization: {
        minimize: false,
        splitChunks: {
            chunks: 'all',
            maxInitialRequests: Infinity,
            minSize: 0,
            cacheGroups: {
                node_modules: {
                    test: /[\\/]node_modules[\\/]/,
                    name: 'vendor'
                }
            },
        },
    },
    module: {
        rules: [
            {
                test: /\.ts$/,
                exclude: /node_modules/,
                loader: 'ts-loader'
            }
        ]
    }
}
```

`excludeRegex`: `/.*\.map/`

Before:
```
demo-project
│   vendor.js
│   vendor.js.map
│
└───src
    └───handlers
            viewer-request.js
            viewer-request.js.map
```

After:
```
demo-project
│   vendor.js
│
└───src
    └───handlers
            viewer-request.js
```

## Todos:

- [x] Write tests
- [x] Write documentation
- [x] Fix linting errors

P.S. I had to add an additional missing undefined-check in `lib/wpwatch.js` to pass 3 failing unit tests.

### DIFF
--- a/lib/packageModules.js
+++ b/lib/packageModules.js
@@ -27,15 +27,19 @@ function setArtifactPath(funcName, func, artifactPath) {
 
 function zip(directory, name) {
   // Check that files exist to be zipped
-  let files = glob.sync('**', {
+  const globFiles = glob.sync('**', {
     cwd: directory,
     dot: true,
     silent: true,
-    follow: true
+    follow: true,
+    nodir: true
   });
 
-  if (this.configuration.excludeRegex) {
-    files = _.filter(files, f => f.match(this.configuration.excludeRegex) === null);
+  const files = this.configuration.excludeRegex
+    ? _.filter(globFiles, f => f.match(this.configuration.excludeRegex) === null)
+    : globFiles;
+  if (files.length < globFiles.length && this.options.verbose) {
+    this.serverless.cli.log(`Excluded ${globFiles.length - files.length} file(s) based on excludeRegex`);
   }
 
   if (_.isEmpty(files)) {
@@ -49,7 +53,7 @@ function zip(directory, name) {
   this.serverless.utils.writeFileDir(artifactFilePath);
 
   const cwd = directory;
-  const source = '*';
+  const source = files;
   const destination = path.relative(cwd, artifactFilePath);
   const zipArgs = {
     source,
@@ -118,7 +122,9 @@ module.exports = {
           () =>
             this.options.verbose &&
             this.serverless.cli.log(
-              `Zip ${_.isEmpty(entryFunction) ? 'service' : 'function'} (using ${hasNativeZip() ? 'native' : 'node'} method): ${modulePath} [${_.now() - startZip} ms]`
+              `Zip ${_.isEmpty(entryFunction) ? 'service' : 'function'} (using ${
+                hasNativeZip() ? 'native' : 'node'
+              } method): ${modulePath} [${_.now() - startZip} ms]`
             )
         );
     });
--- a/lib/wpwatch.js
+++ b/lib/wpwatch.js
@@ -83,7 +83,7 @@ module.exports = {
         }
 
         process.env.SLS_DEBUG &&
-          this.serverless.cli.log(`Webpack watch invoke: HASH NEW=${stats.hash} CUR=${lastHash}`);
+          this.serverless.cli.log(`Webpack watch invoke: HASH NEW=${stats && stats.hash} CUR=${lastHash}`);
 
         // If the file hash did not change there were no effective code changes detected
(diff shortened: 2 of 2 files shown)


j0k3r:
Thanks for providing that fix.
In fact, it'll also solve #744 because your provide the files list to zip instead of using * ([which does not include dotfiles by default](https://github.com/nfriedly/node-bestzip#dotfiles)).

But after testing the `excludeRegex` it seems to not work at all because the regex is a string and the test is wrong.
In the test `tests/packageModules.test.js` we are defining the regex as a real regex:

```js
        module.configuration = new Configuration({
          webpack: {
            excludeRegex: /.*/
          }
        });
```

So the test is working.
Here is the dumped configuration:

```js
Configuration {
  _config: {
    excludeRegex: /.*/,
    webpackConfig: 'webpack.config.js',
    includeModules: false,
    packager: 'npm',
    packagerOptions: {},
    keepOutputDirectory: false,
    config: null,
    concurrency: 4
  },
  _hasLegacyConfig: false
}
```

But when defining the regex in `serverless.yml` the regex is interpreted as a string (thanks for yaml...).
So the test should define the configuration as:

```js
        module.configuration = new Configuration({
          webpack: {
            excludeRegex: '/.*/'
          }
        });
```

And in that case, test are failing.
Here is the dumped configuration:

```js
Configuration {
  _config: {
    excludeRegex: '/.*/',
    webpackConfig: 'webpack.config.js',
    includeModules: false,
    packager: 'npm',
    packagerOptions: {},
    keepOutputDirectory: false,
    config: null,
    concurrency: 4
  },
  _hasLegacyConfig: false
}
```

I tried to use `RegExp` as a workaround without success yet:

```js
new RegExp(this.configuration.excludeRegex)
```

Do you have an idea on how to fix it?

j0k3r:
As you created your PR on your master branch, I can't force push on it. So I created an other PR, cherry-picked your commit and applying some changes. Go check it #780

l1b3r:
@j0k3r Yeah, sorry I didn't pay enough attention to that. I think you should omit the slashes around the actual regular expression if you declare it as a string in serverless.yml. 

Then you can pass this string into `String.prototype.match` and it will do the automatic conversion of this string to a`RegExp` object under the hood.

So, just define the `excludeRegex` as `".*"` in the rest should work.

I can do a separate PR with these changes if you'd like.

j0k3r:
I finally did it in #780, can you have a look?

pveller:
> when zipping files with bestzip, pass files instead of "*" so that only specified files end up in the zip.

I started getting spawn E2BIG on sls package in bestzip where it shells out (nativeZip()) with a very long command line that I would expect to be very different if it was actually using * instead? All the individual files that serverless need to pack in my list of dependencies go over 2x of allowed `ARG_MAX`.

Downgrade to 5.4.0 solved the problem

PR #808, opened 2021-05-03
Fix: Non node functions are attempted to be packaged

roni-frantchi:
## What did you implement:

Closes --

#579 helped in making webpack **not _compile_** any non-node functions, **BUT** the plugin still attempts to _**package**_ what it think would be webpack's output and fails miserably.  

This PR will make it so that non node runtime packages are also not attempted to be packaged.  

## How did you implement it:
`packageModules.js` will now only package functions set as `entryFunctions` so that any functions that were not not listed as entry points will not attempted to be packaged.  

## How can we verify it:

There's a unit test added, plus you could:  
```yaml
provider:
  name: aws
  runtime: nodejs12.x 

functions:
  node:
    handler: lambda.echo

  py:
    handler: backend.lambda.LambdaFunction
    runtime: python 
```

## Todos:

- [x] Write tests
- [x] Write documentation
- [x] Fix linting errors
- [ ] Enable "Allow edits from maintainers" for this PR ([can't check this as this is an organization PR](https://github.com/isaacs/github/issues/1681))

### DIFF
--- a/lib/packageModules.js
+++ b/lib/packageModules.js
@@ -133,9 +133,7 @@ module.exports = {
 
   copyExistingArtifacts() {
     this.serverless.cli.log('Copying existing artifacts...');
-    // When invoked as a part of `deploy function`,
-    // only function passed with `-f` flag should be processed.
-    const functionNames = this.options.function ? [this.options.function] : this.serverless.service.getAllFunctions();
+    const functionNames = _.map(this.entryFunctions, 'funcName');
 
     // Copy artifacts to package location
     if (isIndividialPackaging.call(this)) {
--- a/lib/packageModules.js
+++ b/lib/packageModules.js
@@ -133,7 +133,14 @@ module.exports = {
 
   copyExistingArtifacts() {
     this.serverless.cli.log('Copying existing artifacts...');
-    const functionNames = _.map(this.entryFunctions, 'funcName');
+    const functionNames = _.filter(
+      this.options.function ? [this.options.function] : this.serverless.service.getAllFunctions(),
+      funcName => {
+        const func = this.serverless.service.getFunction(funcName);
+        const runtime = func.runtime || this.serverless.service.provider.runtime || 'nodejs';
+        return runtime.match(/node/);
+      }
+    );
 
     // Copy artifacts to package location
     if (isIndividialPackaging.call(this)) {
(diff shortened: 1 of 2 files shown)


lib/packageModules.js
```diff
@@ -139,7 +139,12 @@ module.exports = {
 
     // Copy artifacts to package location
     if (isIndividialPackaging.call(this)) {
-      _.forEach(functionNames, funcName => copyArtifactByName.call(this, funcName));
+      const nodeFunctionNames = _.filter(functionNames, funcName => {
+        const func = this.serverless.service.getFunction(funcName);
+        const runtime = func.runtime || this.serverless.service.provider.runtime || 'nodejs';
+        return runtime.match(/node/);
```
j0k3r, line 145: Can you create a separate function to re-use this code from line 217?
For example, you can create that function just below the `getHandlerFile` function:

```js
    const isNodeRuntime = funcName => {
      const func = this.serverless.service.getFunction(funcName);
      const runtime = func.runtime || this.serverless.service.provider.runtime || 'nodejs';

      return runtime.match(/node/);
    };
```

And then, re-use it here (and when we build the `allEntryFunctions` const):
```suggestion
        return isNodeRuntime(funcName);
```

timbodeit: Another idea might be to create a new function, that replaces `this.serverless.service.getAllFunctions()` calls and returns an already filtered list. It could be called something like `getAllNodeFunctions()`.

This would also prove useful to fix another issue I'm currently facing, where serverless-webpack still overrides the artifact path of Java based in packageModules.js:161

lib/packageModules.js
```diff
@@ -139,7 +139,12 @@ module.exports = {
 
     // Copy artifacts to package location
     if (isIndividialPackaging.call(this)) {
-      _.forEach(functionNames, funcName => copyArtifactByName.call(this, funcName));
+      const nodeFunctionNames = _.filter(functionNames, funcName => {
```
timbodeit, line 142: The same `nodeFunctionNames` would also need to be used below in line 165 (previously 160) to avoid serverless-webpack still overriding the artifact path of the given function.

lib/packageModules.js
timbodeit: Another option might be to add something like a `getAllNodeFunctions()` at a higher to replace `this.serverless.service.getAllFunctions()` calls with a filtered version. This could then also be used to provide similar logic in files like `validate.js`.

murilolobato:
There is any workaround for webpack avoid packaging a python function while this is not merged into master?

j0k3r:
Done by https://github.com/serverless-heaven/serverless-webpack/pull/876

Issue #299, opened 2017-12-10
JavaScript heap out of memory


daniel-cottone:
# This is a Bug Report

## Description

I'm in the process of trying to upgrade `serverless-webpack` version from `2.2.3`, where I do not experience the following issue. Our serverless configuration has `package: invididually: true` set, and about 40 functions. When I try to upgrade to a later version of `serverless-webpack` and run `sls webpack`, the build will run for about a minute and then I get the following error:

```
lambda:daniel.cottone $ npm run build

> expert-api-lambda@0.1.0 build /Users/daniel.cottone/Projects/expert-api/lambda
> sls webpack --stage dev

Serverless: Bundling with Webpack...
ts-loader: Using typescript@2.5.2 and /Users/daniel.cottone/Projects/expert-api/lambda/tsconfig.json
ts-loader: Using typescript@2.5.2 and /Users/daniel.cottone/Projects/expert-api/lambda/tsconfig.json
ts-loader: Using typescript@2.5.2 and /Users/daniel.cottone/Projects/expert-api/lambda/tsconfig.json
ts-loader: Using typescript@2.5.2 and /Users/daniel.cottone/Projects/expert-api/lambda/tsconfig.json
ts-loader: Using typescript@2.5.2 and /Users/daniel.cottone/Projects/expert-api/lambda/tsconfig.json
ts-loader: Using typescript@2.5.2 and /Users/daniel.cottone/Projects/expert-api/lambda/tsconfig.json
ts-loader: Using typescript@2.5.2 and /Users/daniel.cottone/Projects/expert-api/lambda/tsconfig.json

<--- Last few GCs --->

[42611:0x104001600]    55964 ms: Mark-sweep 1405.7 (1508.8) -> 1405.7 (1508.8) MB, 1721.0 / 0.0 ms  allocation failure GC in old space requested
[42611:0x104001600]    57889 ms: Mark-sweep 1405.7 (1508.8) -> 1405.5 (1487.3) MB, 1923.4 / 0.0 ms  (+ 0.0 ms in 0 steps since start of marking, biggest step 0.0 ms, walltime since start of marking 1923 ms) last resort 
[42611:0x104001600]    59801 ms: Mark-sweep 1405.5 (1487.3) -> 1405.4 (1486.8) MB, 1903.6 / 0.0 ms  last resort 


<--- JS stacktrace --->

==== JS stack trace =========================================

Security context: 0x37341f01ba79 <JS Object>
[...]

PR #570, opened 2020-04-13
New: Use woker-farm to handle webpack multi compile

soda0289:
## What did you implement:

Memory usage of this plugin is really high when using package individually. This is because we store the webpack stats/results from each function compile until all functions are finished compiling. These stats object contain all of the compiled source code for each module and chunk generated. We don't use many properties from webpack stats and this change adds a new function ` processWebpackStats()` that only extracts the properties we need (cliOutput, errors, outputPath, and externalModules). This greatly reduces the memory footprint during compile and lets the garbage collector clean up a lot more memory.

Lowering the memory improves performance a little bit but there's still a chance of a memory leak if you are using many webpack plugins that do not cleanup properly. To fix this we must ensure memory is cleaned up after every function compile. To do this I added the npm package `worker-farm` that runs each function compile in a separate thread/process. This way we can end the process and clean up the memory after each compilation is done. This also lets us compile multiple functions at once and has a huge performance improvement. Almost 50% faster in some repositories I have tested this on. This is the same pattern that `parallel-webpack` uses to improve performance of webpack multi compile. 

Closes #299 

## How did you implement it:

Refactored the code to create 2 different compilers. Single compiler that uses webpack() in main thread, same as current implementation. Multi compiler uses `worker-farm` to run each webpack compilation in a seperate thread. Since each worker is its own process we can only pass in serializable objects into the thread. This means we can pass webpack config file path or a JSON object that does not have functions or other non serializable objects. We use the V8 serialize function to test if a webpack config can be serialized using the HTML structured clone algorithm. This might be a breaking change but I don't think its possible to include a non serializable object into a serverless config. If it is possible it will throw an error letting the user know they should use webpack config file path instead.

## How can we verify it:

 - Enable package individually
 - Run a build or deploy
 - Ensure every function is compiled correclty

## Todos:

- [X] Write tests
- [ ] Write documentation
- [X] Fix linting errors
- [X] Make sure code coverage hasn't dropped
- [X] Provide verification config / commands / resources
- [X] Enable "Allow edits from maintainers" for this PR
- [X] Update the messages below

***Is this ready for review?:*** YES
***Is it a breaking change?:*** NO

PR #858, opened 2021-06-05
Reduce memory usage by releasing webpack stats objects after compile

janicduplessis:
## What did you implement:

The current implementation keeps the full webpack stats object for all functions. This object is huge and causes very large memory usage when using package: individually and a large number of functions.

The webpack stats object is used for 2 things, outputPath and external modules. Instead of keeping the full stats object around, we calculate the outputPath and external modules after compilation and discard the stats object. Then packageModules and packageExternalModules can use these values directly instead of the stats object.

## How did you implement it:

Set outputPath and externalModules in the compile phase and discard webpack stats object.

## How can we verify it:

Tested this change in a large serverless app.

Before: Memory usage slowly increases up to 7.2gb, builds in ~1500s
After: Memory usage hovers around 1.5gb, builds in ~1000s

## Todos:

- [x] Write tests
- [ ] Write documentation
- [x] Fix linting errors

### DIFF
--- a/lib/compile.js
+++ b/lib/compile.js
@@ -4,6 +4,7 @@ const _ = require('lodash');
 const BbPromise = require('bluebird');
 const webpack = require('webpack');
 const tty = require('tty');
+const isBuiltinModule = require('is-builtin-module');
 
 const defaultStatsConfig = {
   colors: tty.isatty(process.stdout.fd),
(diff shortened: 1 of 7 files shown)


janicduplessis:
Accomplish something similar to #570, but simpler.

Should fix #299

j0k3r:
This looks interesting, I'll try out on some project and let you know!
In the mean time, could you rebase against the master? Thanks 🙏

janicduplessis:
Rebased!

j0k3r (approved):
Looks good to me 👍

PR #876, opened 2021-06-10
Fix packaging for non-node functions

FatalTouch:
## What did you implement:

Closes: --

Continuation of https://github.com/serverless-heaven/serverless-webpack/pull/808

This PR will make it so all non-node runtimes are ignored.

## How did you implement it:

Replaced instances of this.serverless.service.getAllFunctions with a new util function that filters and returns only node functions.

## How can we verify it:

Unit tests are added and updated:

Can be tested with any config that include other runtimes like python,java etc.
## Todos:

- [x] Write tests
- [x] Write documentation
- [x] Fix linting errors


j0k3r (approved):
It looks ok, thanks @FatalTouch 🙏

PR #890, opened 2021-06-22
Add '--disableWebpack' flag for ignoring plugin when running Serverless Offline

### LINKED ISSUE #889: Feature Proposal: Add flag for ignoring plugin when running Serverless Offline
# This is a Feature Proposal

## Description

I propose a command-line flag that allows developers to ignore the plugin altogether when running Serverless Offline.
It is especially useful to speed up **hot reload** when you don't need to bundle stuff locally.
This flag is different from `--no-build` which just skips compilation, and uses the most recent bundle.

## Additional Data

* ***Serverless-Webpack Version you're using***: 5.5.1
* ***Webpack version you're using***: 5.38.1
* ***Serverless Framework Version you're using***: 2.48.0
* ***Operating System***: Ubuntu Linux

domcorvasce:
## What did you implement:

Closes #889

## How did you implement it:

I'm not sure if this is the right way to do it, but I just added a check at the start of the `before:offline:start` hook.
I check whether the new flag is specified. If it is, I just return from the hook, hence avoiding triggering the subsequent hooks.

## How can we verify it:

I added a simple unit test, but you can try it out by removing the `.webpack` folder, starting Serverless Offline specifying the `--disableWebpack` flag, and checking that the folder is not regenerated.

## Todos:

- [x] Write tests
- [x] Write documentation
- [x] Fix linting errors


KillDozerX2:
Umm, why is this not being merged?

vicary:
@j0k3r The implementation looks like it's working as described. 

Sorry for not getting the context here, because we already have `--no-build`, why would users want to skip it for `serverless-offline`?

Is it about handlers written in other languages or pre-built images?

domcorvasce:
> Sorry for not getting the context here, because we already have `--no-build`, why would users want to skip it for `serverless-offline`?

To speed up local development and testing, especially when dealing with big projects where Webpack takes some time to complete.

vicary:
@domcorvasce Just curious, how do you apply local code changes to the serverless offline server? Are you running a standalone webpack process separately?

domcorvasce:
@vicary it's a bit of a hack. I'm using `nodemon` to watch for changes to the lambda functions files and restart Serverless Offline.

vicary:
@domcorvasce I think we all are facing the same situation for quite some time already.

Due to historical reasons, `serverless-webpack` splits entries in the [MultiCompiler](https://webpack.js.org/api/node/#multicompiler) style which greatly decreases compile speed.

Believe it or not, given enough memory for chunks, webpack is able to handle hundreds of entries in minutes. All you have to do is to use `thread-loader` to prevent overloading.

I am a bit hesitant to merge this because you should be able to achieve the same thing with the following codes:

```yaml
# serverless.yml
plugins: ${file(serverless-plugins.js)}
```

```javascript
// serverless-plugins.js

const plugins = [...]; // your other plugins

module.exports = process.env.IS_OFFLINE ? pluins
  : [
      "serverless-webpack",
      ...plugins,
    ];
```

What do you think?

domcorvasce:
@vicary yes, that’s fair. I think we can close this PR then.

PR #900, opened 2021-07-01
Register `serverless` as peer dependency

medikoo:
It's to ensure only compatible versions or Framework are used together with a plugin, and if mismatch occurs, user is informed with meaningful error message.

_This PR is part of Serverless Framework initiative to [curate most popular plugins](https://github.com/serverless/serverless/issues/9025)_

### DIFF
--- a/package.json
+++ b/package.json
@@ -79,6 +79,7 @@
     "sinon-chai": "^3.7.0"
   },
   "peerDependencies": {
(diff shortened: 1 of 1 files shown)


j0k3r (approved):

j0k3r:
Thanks @medikoo

PR #911, opened 2021-07-20
fix(packaging): ensure consistent artifact sha

### LINKED ISSUE #881: different sha256sum for zip file when no code changes
# This is a Bug Report

## Description

I run serverless package twice without any changes in code and config. The I got the two zip files, when I compare them hash and found the hash is different.

serverless itself DO set the access and creation timestamp to 1970.01.01, but this plugin not.

## Additional Data

* ***Serverless-Webpack Version you're using***: 
5.5.1
* ***Webpack version you're using***:
4.42.0
* ***Serverless Framework Version you're using***:
2.23.0
* ***Operating System***:
Linux or MacOS

russell-dot-js:
## What did you implement:

Closes #881

bestzip does not support sorting contents or stripping file creation date or modified date, resulting in an indeterminate archive. This breaks serverless' change detection as the zip file created by serverless-webpack has a different checksum each time it is generated, even if the source did not change.

## How did you implement it:

Copied over serverless' zip method as it cannot be used when imported by plugin (the piece we need is coupled with other functionality we need to bypass).

Disabled the native zip method and bypass bestzip by default, as bestzip is not configurable and does not guarantee ordering of contents inside of zip file.

Added new configuration "useNativeZip" if you still want the old behavior, but use the serverless method instead of bestzip's "nodezip" as fallback.

## How can we verify it:

Examples:
* With `useNativeZip: true` in webpack config of serverless.yml, run `serverless package` twice and compare checksums. You will see they have changed. Alternatively, run `serverless deploy` and you will see that serverless detected a change. 
* Now remove that config to use the new default behavior and repeat - you will get the same checksum each time.

## Todos:

- [X] Write tests
- [X] Write documentation
- [X] Fix linting errors
- [] Enable "Allow edits from maintainers" for this PR


russell-dot-js:
@HyperBrain what do you think?

thetrevdev (approved):
Most the logic came from the serverless function here
https://github.com/serverless/serverless/blob/db67b353c9ed578d7dd334d162efa1ec11fbfa18/lib/plugins/package/lib/zipService.js#L59

j0k3r:
So, looks like adding `bestzip` instead of using `archiver` seems to have added a lot of drawbacks ( for a small time improvement. I saw some people complaining about zipping not working or not as intended. They then rollback to the 5.3.5 version (before switching from archiver to bestzip).

Shouldn't we just rollback the whole bestzip addition to use the old archiver version instead? It'll be slower but it'll works.
Or should we keep both and let user define which one to use?

russell-dot-js:
> So, looks like adding `bestzip` instead of using `archiver` seems to have added a lot of drawbacks ( for a small time improvement. I saw some people complaining about zipping not working or not as intended. They then rollback to the 5.3.5 version (before switching from archiver to bestzip).
> 
> Shouldn't we just rollback the whole bestzip addition to use the old archiver version instead? It'll be slower but it'll works.
> Or should we keep both and let user define which one to use?

j0k3r I didn't want to rock the boat too much and delete @HyperBrain's hard work, but happy to remove bestzip altogether as that was my original implementation prior to putting up PR

j0k3r:
`bestzip` support wasn't added by @HyperBrain. Even if it was, if the feature add more negative feedback than positive, it seems legitimate for me to revert it.
Or maybe, like you did, to change the way it is used by default.

So maybe, we can revert back to `archiver` instead of `bestzip` and add an option to use `bestzip` if possible (ie: `zip` is installed, configuration doesn't goes against, etc.)

russell-dot-js:
> `bestzip` support wasn't added by @HyperBrain. Even if it was, if the feature add more negative feedback than positive, it seems legitimate for me to revert it.
> Or maybe, like you did, to change the way it is used by default.
> 
> So maybe, we can revert back to `archiver` instead of `bestzip` and add an option to use `bestzip` if possible (ie: `zip` is installed, configuration doesn't goes against, etc.)

Got it. LMK what you want, I'm happy to gut it completely.

README.md
```diff
@@ -502,6 +503,22 @@ custom:
 This can be useful, in case you want to upload the source maps to your Error
 reporting system, or just have it available for some post processing.
 
+#### Try native zip functionality
+
+Native zip is much faster than node zip, however the "bestzip" library used lacks
```
remi00, line 508: From my testing it also appears that sometimes it will be very helpful to have the `excludeRegex` setting work. Having files like `.DS_Store` or `.yarn-integrity` removed will make the ZIP files even more stable. Maybe this is the additional argument for using `archiver` in place of `bestzip`, as the former one has the `glob()` method?

remi00 (commented):
I work on resolving serverless/serverless#8666 which should be huge improvement to deploy time on Serverless stacks. It's actually a great news that this very ticket will be solved, as it will keep the implementation on Serverless side much simpler. I just have a suggestion to use `archiver` and keep `excludeRegex` supported.

j0k3r:
Thanks for the hint @remi00!

Regarding what's going on in Serverless about deployment optimization, I think we can just drop `bestzip` all at once.

We can arg about that drop for many reasons:
- ability to easier defined which files can be zipped (because it uses `glob()`)
- support for creation date & modified date (useful to determine if the zip has been changed)
- windows support is not perfect (see #905)

remi00 (changes requested):
I just tested and found another thing that is undeterministic:
`lib/packExternalModules.js:27`
External modules variable here can come in random order. The following change does resolve it:
```javascript
   _.forEach(externalModules.sort(), externalModule => {
```

russell-dot-js:
@thetrevdev and I just wrapped this up, please review @remi00 @j0k3r

lib/packageModules.js
```diff
@@ -26,6 +26,73 @@ function setArtifactPath(funcName, func, artifactPath) {
...
+              zip.append(file.data, {
+                name,
+                mode,
+                date: new Date(0) // necessary to get the same hash when zipping the same content
+              });
+            }
+          );
+
+          return zip.finalize();
+        })
```
remi00, line 65: Both `date: new Date(0)` and `contents.sort` are necessary to get the same hash. I'd mention that in the single block comment, for the future generations :wink:

remi00 (commented):
In general looks good :+1: Maybe just please update the comment.

README.md
```diff
@@ -515,7 +516,6 @@ exampleFunction:
 
 ⚠️ **Note: this will only work if your custom runtime and function are written in JavaScript.
 Make sure you know what you are doing when this option is set to `true`**
-
```
j0k3r, line 518: Could you revert these changes? Thanks

lib/packageModules.js
```diff
@@ -113,7 +155,7 @@ function setServiceArtifactPath(artifactPath) {
   _.set(this.serverless, 'service.package.artifact', artifactPath);
 }
 
-function isIndividialPackaging() {
+function isIndividualPackaging() {
```
j0k3r, line 158: 👍🏼 😄

lib/packagers/yarn.test.js
```diff
@@ -80,10 +78,6 @@ describe('yarn', () => {
       const expectedResult = {
         problems: [],
         dependencies: {
-          bestzip: {
-            version: '2.1.5',
-            dependencies: {}
-          },
```
j0k3r, line 86: Can't you leave this file untouched? I think it has nothing to deal with the purpose of the PR.
That test is only about dependencies not related to building the zip file.

Same for all the `tests/mocks/*.mock.json` files

package.json
```diff
@@ -60,6 +59,7 @@
   "devDependencies": {
     "@babel/core": "^7.16.0",
     "@babel/eslint-parser": "^7.16.0",
+    "archiver": "^5.3.0",
```
j0k3r, line 62: Shouldn't this be in the `dependencies` instead? 🤔

lib/packageModules.js
russell-dot-js: > Both `date: new Date(0)` and `contents.sort` are necessary to get the same hash. I'd mention that in the single block comment, for the future generations 

I could, but my intent was to copy serverless' implementation directly (this is copy-pasted). I don't think it's a huge risk but if anything, making changes would make it more difficult to track changes against serverless' (not that I would personally do this work, but someone might want to)

lib/packageModules.js
vicary: @j0k3r It's all greens, I wonder if we need more coverage?

j0k3r: Well, the whole new zip generation seems to not be tested at all 🤔
https://coveralls.io/builds/43972011/source?filename=lib%2FpackageModules.js#L45

vicary: https://coveralls.io/builds/43972011/source?filename=lib%2FpackageModules.js#L199 calls to `isIndividualPackaging` should be throwing but it's not 🤔🤔🤔

russell-dot-js: I don't have capacity to update this any further. If you are looking for test coverage, check serverless' code where it was copied from. Maybe there are tests that can be copied as well.

The current implementation doesn't work, this one does, and we're already using this fork for ourselves. Hopefully someone will carry the torch to satisfy the reviewers and get this merged.

vicary: @russell-dot-js not a problem at all, please consider this thread resolved.

j0k3r:
Closed in favor of https://github.com/serverless-heaven/serverless-webpack/pull/1018

Issue #950, opened 2021-09-03
bug: Cannot read property '1' of null (not #193)


vicary:
# This is a (Bug Report / ~~Feature Proposal~~)

## Description

For bug reports:
* What went wrong?
When I was doing single deploy, it throws the error `Cannot read property '1' of null` after successful bundling.

* What did you expect should have happened?
It correctly deploys.

* What was the config you used?
    Quite huge and generated from pipeline env vars, but it shouldn't matter. Because it is the module identifier changed.

    The regular expression below does not match the actual module identifier string `'external commonjs "aws-sdk"'`,
    https://github.com/serverless-heaven/serverless-webpack/blob/fa2370d8902f4d5e2ad50ab0c06490b6eb9acf0c/lib/compile.js#L31

    It seems like webpack added the keyword `"commonjs"` in the middle.

* What stacktrace or error message from your provider did you see?
```
  TypeError: Cannot read property '1' of null
      at getExternalModuleName (/Users/vicary/Documents/[REDACTED]/node_modules/serverless-webpack/lib/compile.js:32:61)
      at isExternalModule (/Users/vicary/Documents/[REDACTED]/node_modules/serverless-webpack/lib/compile.js:45:77)
      at getExternalModules (/Users/vicary/Documents/[REDACTED]/node_modules/serverless-webpack/lib/compile.js:76:9)
      at /Users/vicary/Documents/[REDACTED]/node_modules/serverless-webpack/lib/compile.js:103:24
      at arrayMap (/Users/vicary/Documents/[REDACTED]/node_modules/lodash/lodash.js:653:23)
      at Function.map (/Users/vicary/Documents/[REDACTED]/node_modules/lodash/lodash.js:9622:14)
      at /Users/vicary/Documents/[REDACTED]/node_modules/serverless-webpack/lib/compile.js:101:14
      at tryCatcher (/Users/vicary/Documents/[REDACTED]/node_modules/bluebird/js/release/util.js:16:23)
      at Promise._settlePromiseFromHandler (/Users/vicary/Documents/[REDACTED]/node_modules/bluebird/js/release/promise.js:547:31)
      at Promise._settlePromise (/Users/vicary/Documents/[REDACTED]/node_modules/bluebird/js/release/promise.js:604:18)
      at Promise._settlePromise0 (/Users/vicary/Documents/[REDACTED]/node_modules/bluebird/js/release/promise.js:649:10)
[...]

PR #951, opened 2021-09-05
Fixed new module identifier test external format

Phrensoua:
New regex to try to match i.e. `external commonjs "pino"` returned by webpack version ^5.52.0 (or earlier, dunno). Should also work when target is not present. otherwise you get a crash: TypeError: Cannot read property '1' of null

## What did you implement:

Closes #XXXXX

## Todos:

- [ ] Write tests
- [ ] Write documentation
- [ ] Fix linting errors

### DIFF
--- a/lib/compile.js
+++ b/lib/compile.js
@@ -28,7 +28,7 @@ function getStatsLogger(statsConfig, consoleLog) {
 }
 
 function getExternalModuleName(module) {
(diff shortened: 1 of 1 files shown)


j0k3r:
Thanks for your help @Phrensoua.
The bug was reported here #950 and I provided a fix here #953

Issue #825, opened 2021-05-12
serverless-webpack 5.5.0 attempts to run Webpack on container images which are not JavaScript


cnorthwood:
# This is a Bug Report

## Description

For bug reports:
* What went wrong?

serverless-webpack is stopping me from deploying lambda containers as part of my service that are not

* What did you expect should have happened?

it should not have touched them as they're not relevant to webpack

* What was the config you used?

N/A

I assume this is related to https://github.com/serverless-heaven/serverless-webpack/pull/723, which I can see the sense in - however, I can't see a way to say "this isn't a container image that Webpack cares about" and hence ignore that lambda in the service definition. Ideally there'd be an option to opt in/out of either Webpack's involvement in particular functions, or container functions as a whole as not all container functions are guaranteed to be part of the service.

## Additional Data

```
$ /Users/chris/Code/report-support/node_modules/.bin/serverless deploy -v --stage dev
Serverless: Using configuration:
{
  "webpackConfig": "api/webpack.config.js",
  "includeModules": {
    "forceInclude": [
      "pg"
    ]
  },
  "packager": "yarn",
  "keepOutputDirectory": true,
  "packagerOptions": {},
  "concurrency": 12
}
 
 Serverless Error ----------------------------------------
 
  No matching handler found for 'apollo' in 'XXX'. Check your service definition.
 
  Get Support --------------------------------------------
     Docs:          docs.serverless.com
     Bugs:          github.com/serverless/serverless/issues
     Issues:        forum.serverless.com
 
  Your Environment Information ---------------------------
     Operating System:          darwin
     Node Version:              14.16.1
     Framework Version:         2.41.1 (local)
     Plugin Version:            4.6.0
     SDK Version:               4.2.2
     Components Version:        3.9.2
```

j0k3r:
For what I understand you are mixing both container and regular JavaScript function in the same serverless config?
Any chance to share a short config to reproduce?

cnorthwood:
```
service: report-support

provider:
  name: aws
  region: eu-west-2
  runtime: nodejs12.x
  stage: ${opt:stage, 'dev'}
  versionFunctions: false

plugins:
  - serverless-webpack

functions:
  server:
    image:
      uri: ${env:DOCKER_IMAGE}
      entryPoint:
        - /app/node_modules/.bin/aws-lambda-ric
      command:
        - apollo.lambda
    role: LambdaGraphQlRole

  migrate:
    handler: db/migrate.up
[...]

PR #959, opened 2021-09-17
Fixes issue causing remote image strings to not be accepted by the plugin

gwdp:
## What did you implement:

Amend fixes on https://github.com/serverless-heaven/serverless-webpack/issues/825 

Issue introduced with image command detection, preventing simple remote images to be used when in usage with webpack plugin. The use case is custom nodejs images built on the fly with the code and pushed on ECR (implemented and used by Hybridless plugin https://github.com/Hybridless/hybridless).

Serverless documentation also indicates its support. (Please, scroll to the very bottom, last example) https://www.serverless.com/blog/container-support-for-lambda

## How did you implement it:

Prevent inclusion of the 'image' into webpack entries, and also detect the image string before checking for the image name. 

## How can we verify it:

```
service: report-support
provider:
  name: aws
  region: ca-central-1
  runtime: nodejs12.x
  stage: ${opt:stage, 'dev'}
  versionFunctions: false
plugins:
  - serverless-webpack
functions:
  server:
    image: public.ecr.aws/lambda/nodejs:latest
    handler: handler.func
custom:
  webpack:
    webpackConfig: api/webpack.config.js
    packager: yarn
    keepOutputDirectory: true
```

## Todos:

- [X] Write tests
- [X] Write documentation
- [X] Fix linting errors

### DIFF
--- a/README.md
+++ b/README.md
@@ -592,6 +592,15 @@ functions:
         - app.handler2
 ```
 
+If you want to use a remote docker image but still need the webpack process before doing so, you can specify it as indicated below:
+
+```yaml
+# serverless.yml
+functions:
+  myFunction1:
+    image: public.ecr.aws/lambda/nodejs:latest
+```
+
 ## Usage
 
 ### Automatic bundling
--- a/README.md
+++ b/README.md
@@ -592,13 +592,16 @@ functions:
         - app.handler2
 ```
 
-If you want to use a remote docker image but still need the webpack process before doing so, you can specify it as indicated below:
+If you want to use a remote docker image but still need the webpack process before doing so, you can specify it as indicated below. 
+This will include this handler into webpack build but still uses the user defined image as the source image. This case is specially
+usefull for processes that build their images on the go with the webpack built source and deploy to an ECR repository for example.
 
 ```yaml
 # serverless.yml
 functions:
   myFunction1:
     image: public.ecr.aws/lambda/nodejs:latest
+    handler: index.js
 ```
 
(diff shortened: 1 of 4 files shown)


vicary:
If I am reading it correctly, you want to specify `function.image` and `function.handler` at the same time to run the webpack bundle with a custom node runtime.

My interpretation of your quoted documentation [container image support for lambda](https://www.serverless.com/blog/container-support-for-lambda) is to read the handler code inside your Dockerfile for each function. In fact [multi-stage builds](https://docs.docker.com/develop/develop-images/multistage-build/) makes it trivial to include a webpack build step. If you want to share the same webpack config between official nodejs runtimes and custom runtimes, it's as easy as adding a `ADD ...` line in the build stage of your Dockerfile.

While I can see the convenience for built-in support, e.g. deno, the [official reference to serverless.yml](https://www.serverless.com/framework/docs/providers/aws/guide/serverless.yml) explicitly forbids using both at the same time. I don't think working out of spec is doing any good as a plugin.

Closing this for now.

gwdp:
You are right.. Can I make change to allow the image as plain string only? This is fully supported and the plugin should not be throwing with that..

Gabriel

vicary:
Yes, `image` as a string should be allowed to pass through.

Please make a new PR for that.

gwdp:
Ok, will do it..

I guess I overthought it, maybe just getting rid of the last commit will do it. That was the initially proposed PR, thanks for looking it out.

PR #987, opened 2021-11-02
feat(packager): add noInstall option

russell-dot-js:
## What did you implement:

Closes #913

## How did you implement it:

* Add noInstall packagerOption for both npm and yarn
* Update the install methods to skip install if noInstall is true
* Update tests to demonstrate

## Todos:

- [x] Write tests
- [x] Write documentation
- [x] Fix linting errors
- [x] Make sure code coverage hasn't dropped
- [x] Provide verification config / commands / resources
- [x] Enable "Allow edits from maintainers" for this PR
- [x] Update the messages below

***Is this ready for review?:*** YES
***Is it a breaking change?:*** YES

PR #1003, opened 2021-11-19
feat(packager): add noInstall option

### LINKED ISSUE #913: Why does serverless-webpack run yarn install for you?
# This is a Question

## Description

See: https://github.com/serverless-heaven/serverless-webpack/blob/master/lib/packagers/yarn.js

serverless-webpack currently runs yarn install (as well as npm install) for you, depending on packager config. However, I can't help but feel that this is a bit redundant. Wouldn't most CI workflows already have run install prior to running serverless webpack? If not, how would serverless webpack even be installed to be able to run the command itself?

This is an issue particularly because it breaks yarn PnP. Are you open to at least having a config to disable the install altogether?

j0k3r:
Supersede #987 to fix typos.

Fix #913

### DIFF
--- a/lib/packagers/npm.js
+++ b/lib/packagers/npm.js
@@ -103,7 +103,10 @@ class NPM {
     return lockfile;
   }
 
-  static install(cwd) {
+  static install(cwd, packagerOptions) {
+    if (packagerOptions.noInstall) {
+      return BbPromise.resolve();
+    }
     const command = /^win/.test(process.platform) ? 'npm.cmd' : 'npm';
     const args = ['install'];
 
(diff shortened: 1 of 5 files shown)


lib/packagers/npm.js
```diff
@@ -103,7 +103,10 @@ class NPM {
     return lockfile;
   }
 
-  static install(cwd) {
+  static install(cwd, packagerOptions) {
+    if (packagerOptions.noInstall) {
+      return BbPromise.resolve();
+    }
```
vicary, line 109:
```suggestion
    }

```

This is really nitpicking but consistent with `lib/packagers/yarn.js` below.

vicary (approved):
LGTM

PR #1004, opened 2021-11-19
Fix command resolution when watch is involved

medikoo:
When testing the plugin with the Serverless Framework, I've discovered that the plugin finalizes the given processing hook (and in turn command)  while leaving some command-specific logic working in the background.

Such configuration introduces some erroneous side effects.  Framework assumes that command was finalized and pursues some cleanup operations. Command seems to operate without issues only because Framework doesn't imply a hard exit on the process.

It's especially important to have it fixed before v3 of the Framework is released, as it comes with new logging interfaces, and e.g. on command finalization, all progress bars are permanently cleared, leaving this feature no further operable (while it's a valuable feature for watching as configured in this plugin)

### DIFF
--- a/lib/run.js
+++ b/lib/run.js
@@ -21,29 +21,32 @@ module.exports = {
       this.serverless.cli.log(`Enabled polling (${watchOptions.poll} ms)`);
     }
 
-    compiler.watch(watchOptions, (err /*, stats */) => {
-      if (err) {
-        throw err;
-      }
-      // eslint-disable-next-line promise/catch-or-return, promise/no-promise-in-callback
-      BbPromise.try(() => {
-        if (this.originalServicePath) {
-          process.chdir(this.originalServicePath);
-          this.serverless.config.servicePath = this.originalServicePath;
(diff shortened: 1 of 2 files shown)


j0k3r (commented):
To be honest, I never used the `--watch` option.
But this seems fair, thanks for submitting the PR ahead of Serverless v3.
Could you just rebase against the master so the branch will be up to date? Thanks

medikoo:
> Could you just rebase against the master so the branch will be up to date? Thanks

Sure, rebased

j0k3r (approved):

vicary (approved):
Never successfully setup the `--watch` option either, whenever I tried `serverless invoke local --watch` it always fails to run upon code change.

image

Despite the error above, a quick check with `serverless@^3.0.0-pre` proves this change indeed prevents a crash.

I am happy to take this at its current state, in fact any improvements are welcomed.

j0k3r:
@medikoo could again rebase against the master? Thanks

medikoo:
Sure, pushed

PR #1005, opened 2021-11-19
Fix user error reporting

medikoo:
In Serverless Framework all user errors are expected to be reported with `ServerlessError` instance (class accessible at `this.serverless.classes.Error`) otherwise error is assumed as programmatic, and when thrown, whole stack trace is presented.

This patch ensures that those common user errors are reported with `ServerlessError`

### DIFF
--- a/lib/compile.js
+++ b/lib/compile.js
@@ -79,7 +79,7 @@ function getExternalModules({ compilation }) {
   return Array.from(externals);
 }
 
-function webpackCompile(config, logStats) {
+function webpackCompile(config, logStats, ServerlessError) {
   return BbPromise.fromCallback(cb => webpack(config).run(cb)).then(stats => {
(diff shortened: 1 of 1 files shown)


j0k3r:
Does `this.serverless.classes.Error` exist in each Serverless version (1.x & 2.x)?

medikoo:
> Does this.serverless.classes.Error exist in each Serverless version (1.x & 2.x)?

Yes, it's there since v1.0.0: https://github.com/serverless/serverless/blob/v1.0.0/lib/Serverless.js#L42

j0k3r (approved):

vicary (approved):

j0k3r:
@medikoo One more rebase :)

medikoo:
@j0k3r rebased :)

PR #1011, opened 2021-11-22
Fix handling of multiple compilation errors

medikoo:
Currently, when compiling multiple bundles, the first error is reported, and the rest are reported in the air (after the command is reported as finalized).

This patch fixes, so all errors are acknowledged, and that command reports final error when compilations of all bundles finalizes

### DIFF
--- a/lib/compile.js
+++ b/lib/compile.js
@@ -87,7 +87,7 @@ function webpackCompile(config, logStats, ServerlessError) {
     _.forEach(stats, compileStats => {
       logStats(compileStats);
       if (compileStats.hasErrors()) {
-        throw new ServerlessError('Webpack compilation error, see stats above');
+        throw _.assign(new ServerlessError('Webpack compilation error, see stats above'), { stats: compileStats });
       }
     });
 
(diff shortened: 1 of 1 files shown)


lib/compile.js
```diff
@@ -99,9 +99,22 @@ function webpackCompile(config, logStats, ServerlessError) {
...
+    configs,
+    config =>
+      webpackCompile(config, logStats, ServerlessError).catch(error => {
+        errors.push(error);
+        return error.stats;
+      }),
+    { concurrency }
+  ).then(stats => {
+    if (errors.length) {
+      if (errors.length === 1) throw errors[0];
```
j0k3r, line 113:
```suggestion
      if (errors.length === 1) {
        throw errors[0];
      }
```

j0k3r (commented):
Any example of the CLI output?

medikoo:
> Any example of the CLI output?

With modern logs for Framework v3 release (which I plan to propose with next PR), without this change, issue exposes as:

Capture d’écran 2021-10-14 à 10 32 59

(Note the another stats error log, after command reported the final error).

After this change, all the stats logged will appear before final command error output

lib/compile.js
medikoo: Thanks, taken

j0k3r (approved):
> which I plan to propose with next PR

How many did you plan to propose? :D
I'm joking, thanks for your contributions!

lib/compile.js
```diff
@@ -87,7 +87,7 @@ function webpackCompile(config, logStats, ServerlessError) {
     _.forEach(stats, compileStats => {
       logStats(compileStats);
       if (compileStats.hasErrors()) {
-        throw new ServerlessError('Webpack compilation error, see stats above');
+        throw _.assign(new ServerlessError('Webpack compilation error, see stats above'), { stats: compileStats });
```
vicary, line 90: This is not actually a change request, but rather grabbing the chance to discuss with people who cares about the project.

When I am experimenting the reverting of MultiCompile/ combined entries approach, I can't help but think of a complete rewrite of this project with TypeScript. The monkey patching approach brought down from Serverless V1 makes debugging and error tracing a bit tricky in many cases.

The effort is not small though, do you think it's worth doing?

medikoo: @vicary thanks for comment, but do you see this discussion specific to this contribution?

This PR is about grouping all compilation errors into one error, and not maintaining a situation where first one is reported, and others jump out aside (after Framework reports that command processing is finalized)

j0k3r: I think it's more a global discussion not specific to your contribution

vicary: Seeing that you're adding `stats` to the `ServerlessError` object, I just expanded the topic a bit.

But definitely not a blocker for this issue, just general discussion.

vicary (approved):

PR #1013, opened 2021-11-24
Modern logs for Serverless Framework v3

medikoo:
Configure modern logs with provided new logging interface, maintaining compatibility with v2 release of a Framework (this PR ensures no changes to log output happens for that version).

New logs can be tested with `serverless@pre-3` release

_After v3 is released, we may consider dropping support for v2, and cleaning up the code so only modern logs are configured - I can help providing a PR_

### v2 Logs (before)

_success:_
Screenshot 2021-10-13 at 17 49 56

_warning:_
Screenshot 2021-10-13 at 17 50 25

_error:_
v2-error

### v3 Logs (after)

_progress:_
Screenshot 2021-10-13 at 17 50 37

_success:_
Screenshot 2021-10-13 at 17 50 08

_warning:_
v3-warning

_error:_
v3-error

PR #1016, opened 2021-11-30
Mark plugin as Serverless Framework v3 compliant

medikoo:
Plugin as is, will work with Serverless Framework v3 without issues, therefore it'll be great to mark it as compliant (so it doesn't report a lack of compatibility once v3 is released)

For complete adaptation, it'll be great if it also comes with modern logs (addressed at https://github.com/serverless-heaven/serverless-webpack/pull/1013)

### DIFF
--- a/package.json
+++ b/package.json
@@ -79,7 +79,7 @@
     "sinon-chai": "^3.7.0"
   },
   "peerDependencies": {
-    "serverless": "1 || 2",
+    "serverless": "1 || 2 || 3",
     "webpack": ">= 3.0.0 < 6"
   },
   "optionalDependencies": {


j0k3r (approved):

medikoo:
@j0k3r when we can expect it to be published? We're aiming to release v3 soon, and without it users will be informed that plugin is not compatible with a new version

j0k3r:
Sadly there is no so much we can do.
https://github.com/serverless-heaven/serverless-webpack/pull/1013 is blocked because coverage dropped (we understand why and it's ok) but it's required to merge.
I can't change it, only @HyperBrain have the right to do it.
I've requested him to remove that requirement so we can merge but so far, he's pretty silent.

medikoo:
@j0k3r can we have it PR published without #1013?

This plugin is perfectly runnable in v3 without new logs, and not publishing bumped version range in `peerDependencies` will leave false information to users that the plugin is not compatible, which is not the case.

j0k3r:
All right, I'll release 5.6.1!

PR #1018, opened 2021-12-02
Ensure consistent artifact `sha` (remove `bestzip` and bring back `archiver`)

### LINKED ISSUE #881: different sha256sum for zip file when no code changes

j0k3r:
Replace #911

Closes #881


vicary (approved):
LGTM

Dizolivemint:
just curious if there are any plans to add test coverage and merge @vicary @j0k3r

j0k3r:
We tried to add e2e tests but we failed for now.

lib/packageModules.js
```diff
@@ -26,7 +26,82 @@ function setArtifactPath(funcName, func, artifactPath) {
   }
 }
 
-function zip(directory, name) {
+/**
+ * Copy pasted from Serverless
+ *
+ * @see https://github.com/serverless/serverless/blob/63d54e1537e10ae63c171892edd886f6b81e83f6/lib/plugins/package/lib/zipService.js#L65
+ */
+function serverlessZip(args) {
```
miguel-a-calles-mba, line 34: It seems the code coverage dropped because there are no unit tests for this new function.

j0k3r: It's hard to test unit it, in fact, we would like to provide e2e tests which will more easy to cover that code.

PR #1019, opened 2021-12-02
Add integration tests

j0k3r:
Bring the logic from [Serverless](https://github.com/serverless/serverless/tree/79b5b5e8a91f95f82140c77c113504c7b095507c/test/integrationPackage).

For now there are only two tests:
- packaging a simple application with one deps (and little config)
- packaging a simple application with one deps with `keepOutputDirectory` enabled

The idea is to have one dummy project in the repo.
Before the integration test, that project is copied in a temporary folder. Then deps are installed (it tooks around 1 or 2 min).

After that, each test runs `serverless package` and then check about generated files / zip.
Each test can modify the default serverless configuration.


vicary:
Let me take a deeper look in the morning.

In fact I tried `runServerless` in https://github.com/serverless/serverless/pull/10177, it mocks lifecycle events pretty well. Do you think it's worth using it here?

j0k3r:
@vicary interesting, I'll take a look

PR #1062, opened 2022-02-03
Fix tests with Serverless v3

j0k3r:
The new serverless class must have 3 options for its argument.

```js
// before
serverless = new Serverless();

// after
serverless = new Serverless({ commands: ['print'], options: {}, serviceDir: null });
```

Also:
- remove useless `prettier-eslint-cli` which isn't compatible with ESLint 8 and also isn't necessary (we can use `prettier` directly)
- move all `*.tests.js` file into the `tests/` folder (easier to find them)
- add node 16 to test build
- add build with Serverless v2
- run `npm run format` (that's there is a lot of small changes)

See https://github.com/serverless-heaven/serverless-webpack/pull/1061


j0k3r:
@medikoo I'm wondering why `serverless = new Serverless();` doesn't work with v3.
I must define some default parameters (found [here](https://github.com/serverless/serverless/blob/main/test/unit/lib/serverless.test.js#L25)) not to have errors.
```js
serverless = new Serverless({ commands: ['print'], options: {}, serviceDir: null });
```

Shouldn't these arguments be by default?

Here is the error:

```
     TypeError: undefined is not an array
      at module.exports (node_modules/type/lib/resolve-exception.js:12:14)
      at module.exports (node_modules/type/array/ensure.js:15:25)
      at new Serverless (node_modules/serverless/lib/serverless.js:84:22)
      at Context.<anonymous> (lib/prepareOfflineInvoke.test.js:29:18)
      at processImmediate (internal/timers.js:461:21)
```

medikoo:
> Shouldn't these arguments be by default?

Not really, as there's no default _command_ assumed. Maybe _options_ could be optional, but it shouldn't make much difference.

Starting with v3 `Serverless` is CLI agnostic, and while in v2 it resolved internally _command_ and _options_ from `processs.argv`. Starting from v3 it doesn't assume anything from `process.argv` but requires this data to be provided to the constructor. It's to make programmatic use more friendly (e.g. it may allow deploying to different services at once in the context of the same Node.js process)

j0k3r:
Thanks for the explanation!

package.json
```diff
@@ -28,16 +28,14 @@
   },
   "homepage": "https://github.com/serverless-heaven/serverless-webpack#readme",
   "scripts": {
-    "test": "mocha tests/all index.test.js \"lib/**/*.test.js\" -R spec --recursive",
-    "test-coverage": "nyc ./node_modules/mocha/bin/_mocha tests/all index.test.js \"lib/**/*.test.js\" -R spec --recursive",
-    "eslint": "eslint --ext .js *.js lib tests",
-    "format": "prettier-eslint \"$(pwd)/**/*.js\" --write"
```
vicary, line 34: Does deleting `npm run format` means it is being replacing by `npm run eslint -- --fix` in the future?

vicary (approved):
Just a clarifying question, everything else lgtm.

package.json
j0k3r: Well not in the future, it can already be using like that.
Do you want a `"eslint-fix": "eslint --ext .js *.js lib tests --fix",` script command?
And formatting will now make the `lint` build to fail too.

vicary: > Do you want a "eslint-fix": "eslint --ext .js *.js lib tests --fix", script command?

No I think it's good enough already, just want to make sure I understand the change.

Issue #1140, opened 2022-04-22
Cannot read property 'serverless' of undefined


rizowski:
# This is a Bug Report

## Description
When packaging or deploying it fails to find serverless. I replicated the issue with latest versions of serverless-webpack. If I revert 1 version behind the latest 5.7.0 to 5.6.1 it builds fine with the dependencies I have installed.

For bug reports:
* What went wrong?
When running `yarn sls package` - or deployment doesn't matter - I receive this error.
```
TypeError: Cannot read property 'serverless' of undefined
    at /../../node_modules/serverless-webpack/lib/packageModules.js:102:22
```
* What did you expect should have happened?
Build my project
* What was the config you used?
```js
const path = require("path");
const nodeExternals = require("webpack-node-externals");
const slsw = require("serverless-webpack");
const ForkTsCheckerWebpackPlugin = require("fork-ts-checker-webpack-plugin");

const isLocal = slsw.lib.webpack.isLocal;

module.exports = {
  entry: slsw.lib.entries,
  target: "node14",
  node: {
    __dirname: true
  },
  devtool: isLocal ? "inline-source-map" : undefined,
  externalsPresets: { node: true },
  externals: [
    nodeExternals({
      modulesFromFile: true
    })
  ],
  mode: isLocal ? "development" : "production",
  resolve: {
    extensions: [".ts", ".tsx", ".js", ".json"]
  },
  module: {
    rules: [
      {
        test: /\.tsx?$/,
        loader: "ts-loader",
        exclude: /node_modules/,
        options: {
          transpileOnly: true
        }
      },
      { test: /\.graphql?$/, loader: "@graphql-tools/webpack-loader" }
    ]
  },
  plugins: [new ForkTsCheckerWebpackPlugin({ typescript: { memoryLimit: 8192 } })],
  output: {
    libraryTarget: "commonjs",
    path: path.join(__dirname, ".webpack"),
    filename: "[name].js"
  }
};

```
* What stacktrace or error message from your provider did you see?
```
Error:
TypeError: Cannot read property 'serverless' of undefined
    at /../../my-project/node_modules/serverless-webpack/lib/packageModules.js:102:22
    at tryCatcher (/../../my-project/node_modules/bluebird/js/release/util.js:16:23)
    at Promise._settlePromiseFromHandler (/../../my-project/node_modules/bluebird/js/release/promise.js:547:31)
[...]

PR #1153, opened 2022-05-09
Fix `this.serverless` is undefined during zip

j0k3r:
Calling `serverlessZip` directly does not _inject_ (or hydrate) the `this` inside the `serverlessZip`.
We need to use `.call` for that.

Fix https://github.com/serverless-heaven/serverless-webpack/issues/1140

### DIFF
--- a/lib/packageModules.js
+++ b/lib/packageModules.js
@@ -143,7 +143,7 @@ function zip(directory, zipFileName) {
   const artifactFilePath = path.join(this.webpackOutputPath, zipFileName);
   this.serverless.utils.writeFileDir(artifactFilePath);
 
-  return serverlessZip({
+  return serverlessZip.call(this, {
     directory,
     artifactFilePath,
     files


lib/packageModules.js
```diff
@@ -143,11 +143,15 @@ function zip(directory, zipFileName) {
...
-  });
+  return serverlessZip
+    .call(this, {
+      directory,
+      artifactFilePath,
+      files
+    })
+    .then(zipPath => {
+      return zipPath;
+    });
```
vicary, line 154: Is this necessary? 🤔

j0k3r: Looks like not

vicary (approved):
LGTM, let's see if @3alqassab @rizowski has follow ups on #1140.

rizowski:
This fixed my initial issue. I am receiving a different error but I don't believe it is related. LGTM

PR #1186, opened 2022-06-28
Match NPM `--ignore-scripts` w/ Yarn Support

hnryjms:
## What did you implement:

Allow NPM packager to use `--ignore-scripts` option to match Yarn behavior. This is useful for skipping build/compile steps that are irrelevant because AWS Lambda is a different CPU architecture than the build machine.

## How did you implement it:

Copied implementation from Yarn.

## How can we verify it:

```yml
custom:
  webpack:
    packagerOptions:
      ignoreScripts: true
```


vicary (approved):
LGTM

PR #1187, opened 2022-06-30
Define Serverless Webpack CLI options on other commands

### LINKED ISSUE #1121: `--watch` flag not working with serverless v3
Hi, first thanks for your awesome plugin. Updating to latest serverless in the 3.x series (v3.12.0) with latest serverless-webpack (5.6.1), I noticed `--watch` stopped working. It previously worked flawlessly using serverless 2.x.

When I run:

`serverless invoke local --function myfunction --path event.json --watch`

I then get error: 

`Detected unrecognized CLI options: "--watch".`

Having the `--watch` option work is important for productivity and great dev experience. If I don't provide a `--watch`, I'm able to successfully invoke my local function. Any ideas what might be the issue with `--watch` not working?

Thanks!
George

## Additional Data

* ***Serverless-Webpack Version you're using***: 5.6.1
* ***Webpack version you're using***: 5.71.0
* ***Serverless Framework Version you're using***: 3.12.0
* ***Operating System***:  Apple M1 (macOS Monterey 12.2.1)

`serverless invoke local --function myfunction --path event.json --watch`

Screen Shot 2022-04-06 at 9 28 17 PM

j0k3r:
This is required with Serverless v3: each CLI options must be declared.
As we are using some options only on the context of commands defined in other plugin, we need to defined them.

This brings back these commands:
- `serverless invoke local --watch`
- `serverless invoke local --webpack-use-polling=4000`
- `serverless offline --webpack-no-watch`
- `serverless offline start --webpack-no-watch`

Also:
- `dontPrintOutput` has been replaced by `printOutput` in ServerlessOffline plugin which force the output, so we can remove that information from the README

Fix #1121

@vicary should we re-introduce the `--no-build` CLI option as it seems to be useful not to have the `noBuild` hardcoded during CI (for example)?


vicary:
@j0k3r I think so. `--no-build` is useful for automated environment setups, e.g. ansible and/or docker image building.

j0k3r:
@vicary Great, I'll add it back in an other PR.
I'm wondering why did you use the CLI option `build` instead of `no-build` here?
https://github.com/serverless-heaven/serverless-webpack/blob/aa89ac6058b01cdd26d2b9e9482a990eabcfa3ae/lib/prepareOfflineInvoke.js#L13

vicary:
@j0k3r I did not invent it.

Never dig too deep, I just assume severless will automatically convert `no-` prefix to a false for that name.

vicary:
The idea comes from [this commit](https://github.com/serverless-heaven/serverless-webpack/commit/44dc73e124358bdb2308963454498517289a12ee#diff-cabded08ff9373812a499e53913ccaca1d50f6d981eefa5c45f8cb88809a9cf7L195-L197).

If the `no-` prefix is still treated the same way in V3, [reading ['no-build']](https://github.com/serverless-heaven/serverless-webpack/pull/1190/files#diff-f2a6b385e8c08af8a3a569cd71a05e24b6a95af603dc3b5e2ee47c201df33c49R1200) may not work. I think it's worth double checking with an actual CLI run.

j0k3r:
@vicary good for review.
The `--no-build` stuff is handled in [a different PR](https://github.com/serverless-heaven/serverless-webpack/pull/1190).

vicary (approved):
LGTM!

PR #1007, opened 2021-11-19
feat(compiler): add noBuild option

vicary:
## What did you implement:

Closes #965

I removed the `--no-build` CLI option from README to discourage usage, the actual feature is kept for backward compatibility.

## How did you implement it:

```yml
custom:
  webpack:
    noBuild: true
```

## How can we verify it:

```bash
> serverless deploy
> serverless invoke local
> serverless offline
```

## Todos:

- [x] Write tests
- [x] Write documentation
- [x] Fix linting errors
- [x] Make sure code coverage hasn't dropped
- [x] Provide verification config / commands / resources
- [x] Enable "Allow edits from maintainers" for this PR
- [x] Update the messages below

***Is this ready for review?:*** YES
***Is it a breaking change?:*** NO

PR #1190, opened 2022-07-04
Add the `--skip-build` CLI option 

### LINKED ISSUE #965: local invoke with --no-build flag(s) causes deprecation warnings
# This is a (Bug Report)

## Description

Using --no-build (and --webpack-offline) causes deprecation warnings

`serverless invoke local --no-build --webpack-offline -f ${FUNCTION}`

```
Detected unrecognized CLI options: "--build", "--webpack-offline".
Starting with the next major, Serverless Framework will report them with a thrown error
More Info: https://www.serverless.com/framework/docs/deprecations/#UNSUPPORTED_CLI_OPTIONS
```

## Additional Data

    "serverless": "^2.44.0",
    "serverless-offline": "8.2.0",
    "serverless-webpack": "^5.5.0",
    "webpack": "^5.4.0"

j0k3r:
We removed it because it was generating some deprecated message and we introduced the parameter `noBuild` instead.
Now the CLI option is back and the parameter is kept.

Kind of revert https://github.com/serverless-heaven/serverless-webpack/pull/1007

Double fix #965

The first commit is from https://github.com/serverless-heaven/serverless-webpack/pull/1187, it'll be dropped once the PR is merged and that one is rebased.

### DIFF
--- a/.lintstagedrc.yml
+++ b/.lintstagedrc.yml
@@ -1,4 +1,2 @@
 "*.js":
-  - "prettier-eslint --write"             # Run Prettier
-  - "eslint"                              # Run TSLint
-  - "git add"
+  - "eslint"
--- a/tests/index.test.js
+++ b/tests/index.test.js
@@ -219,7 +219,7 @@ describe('ServerlessWebpack', () => {
             });
 
             it('should skip compile if requested', () => {
-              slsw.options.build = false;
+              slsw.options['skip-build'] = false;
               slsw.skipCompile = true;
               return expect(slsw.hooks['before:invoke:local:invoke']())
                 .resolves.toBeUndefined()
@@ -377,8 +377,7 @@ describe('ServerlessWebpack', () => {
           name: 'before:offline:start',
           test: () => {
             it('should prepare offline', () => {
-              slsw.skipCompile = false;
-              slsw.options.build = true;
+              slsw.options['skip-build'] = true;
               return expect(slsw.hooks['before:offline:start']())
                 .resolves.toBeUndefined()
                 .then(() => {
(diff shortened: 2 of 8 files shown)


tests/index.test.js
```diff
@@ -418,7 +418,7 @@ describe('ServerlessWebpack', () => {
                 });
             });
             it('should skip compiling when requested', () => {
-              slsw.skipCompile = true;
+              slsw.skipCompile = false;
```
vicary, line 421: Something doesn't add up here, did we flip the boolean meaning in `validate.js`?

j0k3r: I've just reverted what you did here: https://github.com/serverless-heaven/serverless-webpack/commit/44dc73e124358bdb2308963454498517289a12ee#diff-59e25b254be93038f106111be580b6fb54c6963b6c4e7ef744e58fb8f2b3e3a2

j0k3r: In fact, it changes nothing if it's true or false because the value is updated by the new option. I'll just remove the line.

j0k3r:
Because of Serverless automatically handling `--no-build` option (and we should then declare only the `--build` option) I prefer to use `--skip-build` because we don't need the `--build` AND `no-build` option. It confuses me.

.lintstagedrc.yml
```diff
@@ -1,4 +1,2 @@
 "*.js":
-  - "prettier-eslint --write"             # Run Prettier
```
vicary, line 2: @j0k3r Does it mean it won't autofix from now on, but still prevents commit on lint errors?

vicary (approved):

PR #1191, opened 2022-07-04
fix: `packagerOption` lockFile is now properly used (for NPM)

moroine:
**Changes:**
- `packagerOptions.lockFile` need to be passed to packager (useful for npm workspace where the lock file is in the root level).

### DIFF
--- a/lib/packExternalModules.js
+++ b/lib/packExternalModules.js
@@ -283,225 +283,231 @@ module.exports = {
         this.options.verbose && this.serverless.cli.log(`Fetch dependency graph from ${packageJsonPath}`);
       }
 
-      return packager.getProdDependencies(path.dirname(packageJsonPath), 1).then(dependencyGraph => {
-        const problems = _.get(dependencyGraph, 'problems', []);
-        if (this.options.verbose && !_.isEmpty(problems)) {
-          if (this.log) {
-            this.log.verbose(`Ignoring ${_.size(problems)} NPM errors:`);
-          } else {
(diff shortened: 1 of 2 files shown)


j0k3r (commented):
Just to be sure what you changed (because looks like the whole Git diff is unreadable), you added `this.configuration.packagerOptions` as the 3rd parameter to `packager.getProdDependencies`? Nothing else?

moroine:
> Just to be sure what you changed (because looks like the whole Git diff is unreadable), you added `this.configuration.packagerOptions` as the 3rd parameter to `packager.getProdDependencies`? Nothing else?

Yes, it was missing from my PR. I've also change the `packageLockPath` to use packagerOptions when provided

j0k3r (approved):

vicary:
Ah, I have done that before.

It's tempting to format the files when working on it, but it makes reviews harder than it should.

In fact it's fine with a seperate chore PR that re-formats every single file.

@moroine Do you think it's possible to follow the existing styles, so your changes would stand out in the diffs?

j0k3r:
I think adding the extra 3rd parameter make the line too long and it's then formatted by prettier.

vicary:
True! Even github.dev showed the diff way better than the Web UI, my fault for reading it wrong.

vicary (approved):

moroine:
> True! Even github.dev showed the diff way better than the Web UI, my fault for reading it wrong.

Yup, the eslint re-format the file 😢  

@vicary @j0k3r  the setting you're looking for is the `hide whitespace` 
image

Issue #548, opened 2020-01-10
support layer-provided lambda handlers (e.g. new relic lambda layer)


thekelvinliu:
# This is a Feature Proposal

## Description

i'm using this plugin to keep my lambda deployments small and fast -- it's fantastic, and i really appreciate the effort put into this project 😄 i'd like to instrument my nodejs lambda with newrelic, using their published lambda layer. i am using their [manual steps documentation](https://docs.newrelic.com/docs/serverless-function-monitoring/aws-lambda-monitoring/get-started/enable-new-relic-monitoring-aws-lambda#manual-layer). this requires changing the handlers in serverless config to `newrelic-lambda-wrapper.handler` and setting `NEW_RELIC_LAMBDA_HANDLER` to the original handler path. this breaks serverless-webpack validation:

```
No matching handler found for 'newrelic-lambda-wrapper' in '/path/to/repo'. Check your service definition.
```

in order to support this type of setup, this plugin would need more flexibility with the handlers. instead of only looking at the function config, maybe there could be other ways of setting the handler path? for example in my case, i'd want to configure it to use `NEW_RELIC_LAMBDA_HANDLER` environment variable set on the function config.

## Additional Data

* ***Serverless-Webpack Version you're using***: 5.3.1
* ***Webpack version you're using***: 4.41.5
* ***Serverless Framework Version you're using***: 1.60.5
* ***Operating System***: macOS 10.14.6
* ***Stack Trace (if available)***:

proddimus:
This has caused quite a bit of frustration, great to get this fixed.

kirrg001:
+1

Running into this issue because the handler is inside the lambda layer. This is working great with `serverless` and `serverless-offline`, but not with `serverless-webpack`.

Do you a accept a PR to skip this validation step?

PR #1220, opened 2022-08-16
feat: function level entry overrides

kirrg001:
## What did you implement:

refs #548

Add support for layer provided handler.

## How did you implement it:

`serverless-webpack` is heavily relying on the existence of the defined lambda handler fn.
This won't work for lambda layer provided handlers.

The easiest fix I could think of is to add a custom lambda entrypoint configuration, which is used alternatively.

The remote lambda handler fn is located inside the layer therefor webpack cannot find the handler. 

With this fix the logic will ignore that and use the local custom handler instead.

In general: Layer provided handlers will call the local/original handler. The remote layer handler acts as a wrapper. This is how it works for NewRelic or Instana.

**This fix was only tested with remote ARN's. This is not tested with local layers. But I assume it works the same. I would like to kindly ask to ignore this for now.**

## How can we verify it:

```
service: myservice

provider:
  name: aws
  runtime: nodejs12.x
  stage: dev
  region: us-east-2

plugins:
  - serverless-webpack

custom:
  webpack:
    webpackConfig: ./webpack.config.js
    includeModules: true
    lambdaEntrypoint: 'index.handler'

package:
  individually: true

functions:
  with-serverless-webpack:
    layers:
      - LAYER-ARN-CONTAINING-THE-HANDLER
    handler: lambdalayer.handler
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,17 +71,6 @@ custom:
     excludeFiles: src/**/*.test.js # Provide a glob for files to ignore
 ```
 
-### Layers
-
-```yaml
-functions:
-  my-function:
-    layers:
-      - LAYER-ARN
-    handler: layer.handler
-    entrypoint: src/index.handler
-```
-
 ### Webpack configuration file
 
 By default the plugin will look for a `webpack.config.js` in the service directory. Alternatively, you can specify a different file or configuration in `serverless.yml`.
@@ -196,6 +185,23 @@ module.exports = {
   }, slsw.lib.entries),
   ...
 };
+```
+
+##### Optional entry overrides
+
+`serverless-webpack` generates Webpack entries from the `handler` value by default.
+
+If your handler is different from the webpack entry, e.g. provided by a layer, 
+you may override the generated entry at function level via the `entrypoint` 
+option in `serverless.yml`.
+
+```yaml
+functions:
+  my-function:
+    layers:
+      - LAYER-ARN
+    handler: layer.handler
+    entrypoint: src/index.handler
 ```
 
 #### Full customization (for experts)
--- a/README.md
+++ b/README.md
@@ -69,7 +69,17 @@ custom:
     includeModules: false # Node modules configuration for packaging
     packager: 'npm' # Packager that will be used to package your external modules
     excludeFiles: src/**/*.test.js # Provide a glob for files to ignore
-    lambdaEntrypoint: null # Provide a custom lambda entrypoint. Useful when using layer provided handlers.
+```
+
+### Layers
+
+```yaml
+functions:
+  my-function:
+    layers:
+      - LAYER-ARN
+    handler: layer.handler
+    entrypoint: src/index.handler
 ```
 
 ### Webpack configuration file
--- a/README.md
+++ b/README.md
@@ -69,6 +69,7 @@ custom:
     includeModules: false # Node modules configuration for packaging
     packager: 'npm' # Packager that will be used to package your external modules
     excludeFiles: src/**/*.test.js # Provide a glob for files to ignore
+    lambdaEntrypoint: null # Provide a custom lambda entrypoint. Useful when using layer provided handlers.
 ```
 
 ### Webpack configuration file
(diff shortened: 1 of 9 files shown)


vicary:
@kirrg001 Thanks for the contribution.

From your description I can vaguely imagine the following wrapper:

```ts
// newrelic is provided via /nodejs/node_modules/newrelic in a layer
export { handler } from "newrelic";
```

But it should work without any changes so I must be wrong.

Would you mind showing some minimal codes on how it is supposed to work?

kirrg001:
Hey @vicary!

[here](https://github.com/instana/nodejs/tree/main/packages/aws-lambda-auto-wrap) is an example package.

This package is delivered inside the layer under `nodejs/node_modules/instana-aws-lambda-auto-wrap`.

The lambda handler fn in serverless.yml is defined with `instana-aws-lambda-auto-wrap.handler`.
This wrapper will then call the original handler. Which is e.g. `src/index.js` exposing the "original handler fn".

For both Instana and NewRelic you need to define the original handler path as environment variable e.g. `NEW_RELIC_LAMBDA_HANDLER` or `LAMBDA_HANDLER`.

I hope that helps!

vicary:
@kirrg001 Would this work as a workaround?

## serverless.yml
```yaml
functions:
  with-serverless-webpack:
    environment:
      LAMBDA_HANDLER: index.handler
    layers:
      - instana-auto-wrap-layer-arn
    handler: index.wrapper
```

## index.js
```js
// Exports the wrapper
export.wrapper = require("instana-aws-lambda-auto-wrap").handler;

// The "original handler fn"
export.handler = (...) => {...};
```

## webpack.config.js
```js
module.exports = {
  // Make sure webpack doesn't complain about the wrapper require()
  externals: "instana-aws-lambda-auto-wrap"
};
```

No worries this is not a rejection, just want to make sure I understand the use case.

kirrg001:
> Would this work as a workaround?

Yeah in theory yes. But just in theory ( :

>  just want to make sure I understand the use case.

Yeah you did!

vicary:
I gave it some thoughts on other use cases.

<details>
<summary>opinion/rant</summary>
A plugin level `lambdaEntrypoint` maybe a bit specific for this one.<br/><br/>From my experience I almost never needed the same entry point for all my functions. If I ever need one, I'll wrap it into a `middy` middleware.<br/><br/>It'd be great if instana/newrelic gives you an init function instead, like how sentry do it.
</details>

How about we enable a function level `entrypoint` instead? It then reads the override from `functionDefinition` instead of `this.configuration`.

## serverless.yml
```yml
function:
  with-serverless-webpack:
    entrypoint: src/index.js
    environment:
      LAMBDA_HANDLER: src/index.handler
    handler: nodejs/node_modules/newrelic/src/index.js
    layer:
      - instana-aws-lambda-auto-wrap-arn
```

What do you think?

kirrg001:
Yeah sounds perfect <3

kirrg001:
I am not sure we can add attributes to the function definition because serverless has an inbuilt validation. It might throw a warning

kirrg001:
> Warning: Invalid configuration encountered
  at 'functions.with-serverless-webpack': unrecognized property 'test'
  
That means we also would need to update the serverless codebase. Maybe we should stick with `custom` and `entrypoint`.

vicary:
No worries, it's easy. See [`serverless.configSchemaHandler.defineCustomProperties`](https://www.serverless.com/framework/docs/guides/plugins/custom-configuration#properties-in-custom-via-definecustomproperties), `serverless-plugin-warmup` does it [like this](https://github.com/juanjoDiaz/serverless-plugin-warmup/blob/master/src/schema.js#L137-L152) but we only need a string.

kirrg001:
Great 👍 I will update the PR

kirrg001:
I have updated the change 👍 Please feel free to change/update anything you would like to change (format, docs, code etc)

Happy to get this merged. Thanks for your input & thoughts 🦖

vicary:
Great! It's one test away, we need it the make sure future PRs don't break yours.

I can help with the docs afterwards.

kirrg001:
Ok you want me to add a test? No problem

kirrg001:
Done 👍

vicary (approved):
LGTM. I'll keep it open for a few days in case someone has comment.

miguel-a-calles-mba (approved):

kirrg001:
@vicary Let's merge 🥳

PR #1437, opened 2023-04-12
Update GitHub Actions

j0k3r:
- jump to latest version
- remove deprecated Ubuntu 18.04 and use Ubuntu 20.04 instead, see https://github.blog/changelog/2022-08-09-github-actions-the-ubuntu-18-04-actions-runner-image-is-being-deprecated-and-will-be-removed-by-12-1-22/
- run tests against Node 18 (as AWS Lambda supports it now)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3053,9 +3053,9 @@
       }
     },
     "node_modules/aws-sdk": {
-      "version": "2.1343.0",
-      "resolved": "https://registry.npmjs.org/aws-sdk/-/aws-sdk-2.1343.0.tgz",
-      "integrity": "sha512-p9+bgEeD2HxFMLAl5IyqvxfmFzuzSwyvoKfEaGL3+0Vndztv7TvCbHfKNH6olY9DjyLrq9f34fJ98owGD8sHnQ==",
+      "version": "2.1356.0",
+      "resolved": "https://registry.npmjs.org/aws-sdk/-/aws-sdk-2.1356.0.tgz",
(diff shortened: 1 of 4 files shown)


package-lock.json
```diff
@@ -3044,9 +3044,9 @@
       }
     },
     "node_modules/aws-sdk": {
-      "version": "2.1343.0",
-      "resolved": "https://registry.npmjs.org/aws-sdk/-/aws-sdk-2.1343.0.tgz",
-      "integrity": "sha512-p9+bgEeD2HxFMLAl5IyqvxfmFzuzSwyvoKfEaGL3+0Vndztv7TvCbHfKNH6olY9DjyLrq9f34fJ98owGD8sHnQ==",
+      "version": "2.1356.0",
```
vicary, line 3047: The new v2 may introduce an upgrade warning which is kind of noisy for tests, do we want to supress that?

vicary (approved):

package-lock.json
j0k3r: Oh yeah, it might be a good idea. Let's do the same as in Serverless then

j0k3r: I just checked a random build and it seems there is no noisy warning about the aws-sdk v2 🤷🏼 
https://github.com/serverless-heaven/serverless-webpack/actions/runs/4678437131/jobs/8287167794?pr=1437

vicary: Interesting, serverless/serverless#11753 suggests the warning appears since `aws-sdk@^2.1315` and 2.1356.0 should have it too. But if the warning is not there we can leave it as-is.

PR #381, opened 2018-04-30
Use package json sections defined by packagers

HyperBrain:
## What did you implement:

Closes #379 

## How did you implement it:

The packager implementations now can specify a list of sections in package.json, that will be copied from the original package.json to the ones created for deployment.
For now, only Yarn/resolutions will be added.

## How can we verify it:

Use a project setup like the one mentioned in #379. The generated package.json files in `.webpack/*` should contain a resolution section if one is defined in the project package.json.

## Todos:

- [x] Write tests
- ~~Write documentation~~
- [x] Fix linting errors
- [x] Make sure code coverage hasn't dropped
- [x] Provide verification config / commands / resources
- [x] Enable "Allow edits from maintainers" for this PR
- [x] Update the messages below

***Is this ready for review?:*** YES
***Is it a breaking change?:*** NO

PR #1494, opened 2023-06-07
Allow generating ESM output from npm

### LINKED ISSUE #1493: Allow generating ES Module (ESM) output from npm
# This is a (Bug Report / Feature Proposal)

## Description
- I'd like to generate an ESM output so that I can leverage top-level await and provisioned concurrency to load secrets during lambda initialization, [as described by this AWS blog post](https://aws.amazon.com/blogs/compute/using-node-js-es-modules-and-top-level-await-in-aws-lambda/)
- Here the relevant part of my `serverless.yml`
```yml
provider:
  name: aws
  runtime: nodejs18.x

plugins:
  - serverless-webpack
  - serverless-offline

package:
  individually: true

custom:
  webpack:
    webpackConfig: 'webpack.config.cjs'
    includeModules: true
```
- My `package.json` includes `"type": "module"`
- And my `webpack.config.cjs` (CJS b/c `serverless-webpack` doesn't support an ESM `webpack.config.js`, but that's a separate issue):
```js
const path = require('path');
const nodeExternals = require('webpack-node-externals');

const { isLocal } = slsw.lib.webpack;

module.exports = {
  entry: slsw.lib.entries,
  mode: isLocal ? 'development' : 'production',
  target: 'node18',
  externals: [nodeExternals({ importType: 'node-commonjs' })],
  externalsPresets: { node: true },
  externalsType: 'node-commonjs',
  module: {
    rules: [{
      test: /\.ts$/,
      use: [{ loader: 'ts-loader', options: { transpileOnly: true } }],
      exclude: /node_modules/,
    }],
  },
  resolve: {
    extensions: ['.ts', '.js'],
    modules: ['node_modules', path.resolve(__dirname, './src')],
  },
  experiments: { outputModule: true },
  output: {
    library: { type: 'module' },
    module: true,
  },
};
```
- This works, but it produces `[name].mjs` outputs, [which breaks `serverless-offline`](https://github.com/serverless/serverless/issues/11308) 
- So I set my webpack `output.filename` to `[name].js` and then local / `serverless-offline` works again!
- But now deployed code fails b/c Lambda doesn't know this is an ESM output
- Specifically, the `package.json` generated by `serverless-webpack` is missing the `"type": "module"` that I set as my project default

## Additional Data

* ***Serverless-Webpack Version you're using***: 5.11.0
* ***Webpack version you're using***: 5.85.1
* ***Serverless Framework Version you're using***: 3.32.2
* ***Operating System***: macOS Ventura 13.3.1

aldenquimby:
## What did you implement:

Closes #1493

## How did you implement it:
- #381 appears to be exactly what's needed

## How can we verify it:
- my project works when I include this change locally
- generating `.js` file allows `serverless invoke local` to work, and including `type: module` allows the deployed Lambda to work:
image

## Todos:

- [x] Write tests - `packExternalModules.test.js` already ensures `copyPackageSectionNames` works
- [ ] Write documentation
- [x] Fix linting errors

***Is it a breaking change?:*** YES
- breaking if someone has an ESM package.json today, but wants to produce CJS output
- if this is concerning, an alternative is to push this to a user-defined configuration option: #1495

### DIFF
--- a/tests/packagers/npm.test.js
+++ b/tests/packagers/npm.test.js
@@ -31,7 +31,7 @@ describe('npm', () => {
   });
 
   it('should return no packager sections', () => {
-    expect(npmModule.copyPackageSectionNames).toEqual([]);
+    expect(npmModule.copyPackageSectionNames).toEqual(['type']);
   });
 
   it('requires to copy modules', () => {
--- a/lib/packagers/npm.js
+++ b/lib/packagers/npm.js
@@ -17,7 +17,7 @@ class NPM {
   }
 
   static get copyPackageSectionNames() {
-    return [];
+    return ['type'];
   }
 
   // eslint-disable-next-line lodash/prefer-constant


tests/packagers/npm.test.js
```diff
@@ -31,7 +31,7 @@ describe('npm', () => {
   });
 
   it('should return no packager sections', () => {
-    expect(npmModule.copyPackageSectionNames).toEqual([]);
+    expect(npmModule.copyPackageSectionNames).toEqual(['type']);
```
vicary, line 34: The test name does not reflect your change, please open a new test.

Also if we are talking about ESM, shall we also copy other related sections such as `"main"`, `"exports"`?

I don't have an exhaustive list of sections to be copied, but you get the idea.

aldenquimby: Ok great, I can make those updates

It’s worth noting that this PR assumes that if your source is a module, you want your output to be a module too. This is not the case for many projects, where folks want to develop in ESM and output CJS for compatability. The alternate PR is a larger change, but allows users to opt in to ESM output

vicary: AFAICT copying more sections from package.json should not affect backward compatibility, such that cjs only versions of node.js won't understand esm sections.

But you may want to take a deeper look for potential cases where newer versions of Node.js may incorrectly interpret ESM modules into CJS, this could happen for older modules. My gut feeling is not likely, I'd like your double confirm.

aldenquimby: Looks like the list of fields is: `type`, `exports`, `main`, and `types` (the last two are CJS fallbacks for `exports`). Sources: [node](https://nodejs.org/api/packages.html#nodejs-packagejson-field-definitions) and [ts](https://www.typescriptlang.org/docs/handbook/esm-node.html). `imports` could arguably be included too, but it's internal to a package and if you're using webpack then you shouldn't need `imports` in a build artifact

None of these are safely backwards compatible though. If I copy `type` in my existing project today, without also reworking my webpack config and a few other build tools, I hit this error in Lambda:
```
ReferenceError: require is not defined in ES module scope, you can use import instead

This file is being treated as an ES module because it has a '.js' file extension and '/var/task/package.json' contains \"type\": \"module\". To treat it as a CommonJS script, rename it to use the '.cjs' file extension.
```

You are correct that older versions of Node would work fine, but for anyone on Node 14+, this would be breaking. And I imagine that's most users because Node 12 went EOL in 2020 with security support ending in 2022.

So if we want this to be backwards compatible, I think we should go with the other PR, and I can update the example to copy all 4 of the ESM-related fields

vicary: Makes sense, we can close this and move to the other one.

Issue #1791, opened 2024-05-16
Error running serverless-webpack on Windows with latest Node LTS due to security update


massimocode:
# This is a Bug Report (TLDR at end)

## Description

For bug reports:
* What went wrong? Can't run serverless-webpack on latest LTS Node due to a breaking change in a security update. See: https://github.com/nodejs/node/issues/52554
* What did you expect should have happened? Should run fine
* What was the config you used? Doesn't matter
* What stacktrace or error message from your provider did you see?
```
× Stack api-dev failed to deploy (12s)
Environment: win32, node 20.13.1, framework 3.38.0 (local), plugin 7.2.3, SDK 4.5.1
Credentials: Local, "default" profile
Docs:        docs.serverless.com
Support:     forum.serverless.com
Bugs:        github.com/serverless/serverless/issues

Error:
Error: spawn EINVAL
    at ChildProcess.spawn (node:internal/child_process:421:11)
    at Object.spawn (node:child_process:761:9)
    at childProcess.spawn (C:\REDACTED\api\node_modules\cli-progress-footer\lib\private\cli-progress-footer\disable-props.js:73:50)
    at C:\REDACTED\api\node_modules\serverless-webpack\lib\utils.js:73:32
    at Promise._execute (C:\REDACTED\api\node_modules\bluebird\js\release\debuggability.js:384:9)
    at Promise._resolveFromExecutor (C:\REDACTED\api\node_modules\bluebird\js\release\promise.js:518:18)
    at new Promise (C:\REDACTED\api\node_modules\bluebird\js\release\promise.js:103:10)
    at Object.spawnProcess (C:\REDACTED\api\node_modules\serverless-webpack\lib\utils.js:72:10)
    at NPM.install (C:\REDACTED\api\node_modules\serverless-webpack\lib\packagers\npm.js:143:18)
    at C:\REDACTED\api\node_modules\serverless-webpack\lib\packExternalModules.js:404:20
    at tryCatcher (C:\REDACTED\api\node_modules\bluebird\js\release\util.js:16:23)
    at Promise._settlePromiseFromHandler (C:\REDACTED\api\node_modules\bluebird\js\release\promise.js:547:31)
    at Promise._settlePromise (C:\REDACTED\api\node_modules\bluebird\js\release\promise.js:604:18)
    at Promise._settlePromise0 (C:\REDACTED\api\node_modules\bluebird\js\release\promise.js:649:10)
[...]

PR #1793, opened 2024-05-16
Fix `spawn EINVAL` on Windows when using Node > 20.12.2

j0k3r:
Fix https://github.com/serverless-heaven/serverless-webpack/issues/1791

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -70,7 +70,12 @@ class SpawnError extends Error {
  */
 function spawnProcess(command, args, options) {
   return new BbPromise((resolve, reject) => {
-    const child = childProcess.spawn(command, args, options);
+    const child = childProcess.spawn(command, args, {
+      ...options,
+      // nodejs 20 on windows doesn't allow `.cmd` command to run without `shell: true`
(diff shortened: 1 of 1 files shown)


lib/utils.js
```diff
@@ -70,7 +70,12 @@ class SpawnError extends Error {
  */
 function spawnProcess(command, args, options) {
   return new BbPromise((resolve, reject) => {
-    const child = childProcess.spawn(command, args, options);
+    const child = childProcess.spawn(command, args, {
+      ...options,
```
vicary, line 74: What if we move this line to the end, such that the new `shell` option works like an overridable default?

vicary (changes requested):

lib/utils.js
j0k3r: I asked myself the same question, but we shouldn't be able to let the `shell` option overridable otherwise, we might have some issue on Windows (as we did now)

vicary: That's fair. Given that this is kind of an internal API, we are should be safe to enforce this in case we need to handle more code paths in the future.

vicary (approved):

PR #1794, opened 2024-05-16
Drop Node < 16

j0k3r:
It's no more supported by AWS (creation are blocked since Jan 9, 2024)

https://docs.aws.amazon.com/lambda/latest/dg/lambda-runtimes.html#runtimes-deprecated

I've also updated deps using NPM `10.7.0` (that's why the `lockfileVersion` switched from 2 to 3).


vicary (approved):
May even try 22 since this is not directly tied to Lambda runtimes.

j0k3r:
> May even try 22

22 what? nodejs?

vicary:
Yes, users could build with Node v22 while targeting older Node versions in Lambda (i.e. 16, 18, 20).

Given that minimum additional effort is required for the tests to pass of cause.

Issue #1653, opened 2023-11-14
Error: yarn install --frozen-lockfile --non-interactive failed with code 1


juztinlazaro:
any fix on this? also having an issue in

```
webpack:
  webpackConfig: serverless_v3/webpack.config.js
  includeModules:
    forceInclude:
      - lodash
      - moment
      - es6-promisify
      - axios
      - https
      - axios-retry
      - error-stack-parser
      - stackframe
      - stack-generator
      - stacktrace-gps
      - '@loanmarket/logger-core'
      - '@loanmarket/logger-cloud-watch'
  packager: 'yarn'
```

```
× Stack *****-serverless-sit failed to deploy (6s)
Environment: linux, node 18.12.1, framework 3.36.0 (local), plugin 7.1.0, SDK 4.4.0
Credentials: Local, environment variables
Docs:        docs.serverless.com
Support:     forum.serverless.com
Bugs:        github.com/serverless/serverless/issues

Error:
Error: yarn install --frozen-lockfile --non-interactive failed with code 1
    at ChildProcess.<anonymous> (/home/circleci/project/node_modules/serverless-webpack/lib/utils.js:92:16)
    at ChildProcess.emit (node:events:513:28)
    at ChildProcess.emit (node:domain:489:12)
    at maybeClose (node:internal/child_process:1091:16)
    at ChildProcess._handle.onexit (node:internal/child_process:302:5)

2 deprecations found: run 'serverless doctor' for more details

```

```
const path = require('path');
const slsw = require('serverless-webpack');
const nodeExternals = require('webpack-node-externals');

module.exports = {
  entry: slsw.lib.entries,
  target: 'node',
  mode: slsw.lib.webpack.isLocal ? 'development' : 'production',
  resolve: {
    alias: {
      lib: path.resolve(__dirname, 'lib/'),
      services: path.resolve(__dirname, 'services/'),
      builders: path.resolve(__dirname, 'builders/'),
      constants: path.resolve(__dirname, 'constants/'),
      config: path.resolve(__dirname, 'config/'),
      shared: path.resolve(__dirname, '../shared/'),
    },
  },
  optimization: {
    // We no not want to minimize our code.
    minimize: false,
  },
  performance: {
    // Turn off size warnings for entry points
    hints: false,
  },
  devtool: 'nosources-source-map',
  externals: [nodeExternals()],
  module: {
    rules: [
      {
        test: /\.js$/,
        exclude: /node_modules/,
        use: [
[...]

Issue #1508, opened 2023-06-26
 yarn list --depth=1 --json --production failed with code 1


bebbi:
# This is a (Bug Report / Feature Proposal)

Bug

## Description

For bug reports:
* What went wrong?

When trying to `yarn serverless package --stage=...`,
the error is

` yarn list --depth=1 --json --production failed with code 1`

* What did you expect should have happened?

Do not call `yarn list`or `--production` as these are not supported by `yarn berry`.
For example [here](https://github.com/serverless-heaven/serverless-webpack/blob/aeb9325fdabf70bfd49966b168fd06f60c961be8/lib/packagers/yarn.js#L52)

* What was the config you used?
* What stacktrace or error message from your provider did you see?

```
Error: yarn list --depth=1 --json --production failed with code 1
    at ChildProcess.<anonymous> (/private/tmp/msm-api-new/node_modules/serverless-webpack/lib/utils.js:91:16)
    at ChildProcess.emit (node:events:513:28)
    at ChildProcess.emit (node:domain:489:12)
    at maybeClose (node:internal/child_process:1091:16)
    at Socket.<anonymous> (node:internal/child_process:449:11)
    at Socket.emit (node:events:513:28)
    at Socket.emit (node:domain:489:12)
    at Pipe.<anonymous> (node:net:322:12)
```

For feature proposals:
* What is the use case that should be solved. The more detail you describe this in the easier it is to understand for us.
* If there is additional config how would it look

Similar or dependent issue(s):
* #642

## Additional Data

* ***Serverless-Webpack Version you're using***:
5.11.0
* ***Webpack version you're using***:
5.88.0
* ***Serverless Framework Version you're using***:
3.32.2
* ***Operating System***:

but: yarn 3.4.1

* ***Stack Trace (if available)***:

EhtashamAli:
same issue with npm

vicary:
@bebbi PR is welcomed, you may take reference of #1252 and use the function `isBerryVersion` to give the correct args for yarn 2+ in `getProdDependencies`.

@EhtashamAli `npm ls` is backward compatible for many major versions, please further elaborate the situation in a new issue.

bebbi:
Hi @vicary I've made a quick fix for myself which I'm [happy to share](https://github.com/bebbi/serverless-webpack/compare/master-upstream...fix/berry-support) but it kills yarn classic support.


In order to turn this change into a proper PR, I will need contributor support:

[...]

PR #1803, opened 2024-05-23
When using `SLS_DEBUG` env var, more log will be dumped

kennyhyun:
## What did you implement:

Related to #1508 and probably #1653 too

## Todos:

- [x] Write tests
- [ ] Write documentation
- [x] Fix linting errors

### DIFF
--- a/lib/packagers/npm.js
+++ b/lib/packagers/npm.js
@@ -90,6 +90,9 @@ class NPM {
           if (!failed && !_.isEmpty(err.stdout)) {
             return BbPromise.resolve({ stdout: err.stdout });
           }
+          if (process.env.SLS_DEBUG) {
+            console.error(`DEBUG: ${err.stdout}\nSTDERR: ${err.stderr}`);
+          }
         }
 
         return BbPromise.reject(err);
--- a/lib/packagers/yarn.js
+++ b/lib/packagers/yarn.js
@@ -79,6 +79,9 @@ class Yarn {
           if (!failed && !_.isEmpty(err.stdout)) {
             return BbPromise.resolve({ stdout: err.stdout });
           }
+          if (process.env.SLS_DEBUG) {
+            console.error(`DEBUG: ${err.stdout}\nSTDERR: ${err.stderr}`);
+          }
         }
 
         return BbPromise.reject(err);
--- a/tests/packagers/npm.test.js
+++ b/tests/packagers/npm.test.js
@@ -22,8 +22,16 @@ jest.mock('../../lib/utils', () => {
 });
 
 describe('npm', () => {
+  const ENV = process.env;
+
+  beforeAll(() => {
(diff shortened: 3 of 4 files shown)


vicary:
@kennyhyun What was the errors you are working with?

Depending on the specific error that is sent to `stdout` instead of `stderr`, it could either be
1. An incorrect implementation, or
2. Changing `toString()` in `SpawnError` may be a better option instead.

j0k3r:
@vicary I think it's related to https://github.com/serverless-heaven/serverless-webpack/issues/1508#issuecomment-2126189271

kennyhyun:
yes @vicary thanks

@vicary yeah I tried SpawnError as well but it was not logged.

and `SpawnError` is an exported function, so I could not imagine the consequences.

if it logs errors, and you guys think `SpawnError` is better, I can change

Thanks for the quick response!

kennyhyun:
after reading more, I think npm/yarn packager doesn't log stdout from the SpawnError instance

should I merge stdout into stderr if stderr was empty string or packagers should be updated? I am two minded

kennyhyun:
@vicary one of the errors I was dealing with was,

```
.../@nodert-win10-rs4/windows.storage install$ node-gyp rebuild
../../../../../node_modules/parse-domain postinstall$ run-s build:tries
.../@nodert-win10-rs4/windows.storage install: gyp info it worked if it ends with ok
.../@nodert-win10-rs4/windows.storage install: gyp info using node-gyp@5.1.1
.../@nodert-win10-rs4/windows.storage install: gyp info using node@18.20.2 | linux | x64
.../@nodert-win10-rs4/windows.storage install: gyp info find Python using Python version 3.11.2 found at "/usr/bin/python3"
.../@nodert-win10-rs4/windows.storage install: (node:103869) [DEP0150] DeprecationWarning: Setting process.config is deprecated. In the future the property will be read-only.
.../@nodert-win10-rs4/windows.storage install: (Use `node --trace-deprecation ...` to show where the warning was created)
.../@nodert-win10-rs4/windows.storage install: gyp info spawn /usr/bin/python3
.../@nodert-win10-rs4/windows.storage install: gyp info spawn args [
```

which resolved by downgrading python to 3.10

kennyhyun:
Sorry I cannot reproduce that anymore.

Can I just console.log stdout when SLS_DEBUG was specified instead?

vicary:
@j0k3r thanks for referencing to the source!

@kennyhyun It's not easy to generalize a solution at the level of `spawnProcess`. But when we look at `npm.js` and `yarn.js`, it's more safer to do stdout + stderr at the respective `.catch()`. What do you think?

kennyhyun:
@vicary thanks for the feedback.

yeah I agree, let me do that in `npm.js` and `yarn.js`

vicary (approved):

PR #1287, opened 2022-11-21
adding missing optional peer deps

john-arccos:
## What did you implement:

Updated optional peer dependencies to fix warnings

Closes #XXXXX - didn't make a ticket, pretty straight forward.

This screenshot sums up the issue pretty nicely. `yarn link` is using my changes, `yarn unlink` is using the version from npm:
Screen Shot 2022-11-21 at 3 02 24 PM

## How can we verify it:

If you have a project that's using serverless-webpack typescript, and ts-node, you'll see this warning appear when you install it. If you use my branch instead, the warning goes away.

## Todos:

- [N/A?] Write tests
- [N/A?] Write documentation
- [N/A] Fix linting errors
- [X] Make sure code coverage hasn't dropped
- [X] Provide verification config / commands / resources
- [X] Enable "Allow edits from maintainers" for this PR
- [X] Update the messages below

***Is this ready for review?:*** YES
***Is it a breaking change?:*** It shouldn't be. Added two `*` optional peer dependencies and optional peer `@typescript@>=2.0`, which was already a requirement of ts-node. I guess if they have typescript 1.X but don't have ts-node it would be breaking? Not sure how you want to classify that. Alternatively, changing it to `*` would mean it's definitely non-breaking.

PR #636, opened 2020-09-03
Move `ts-node` as optional dependency

j0k3r:
## What did you implement:

Closes #633

## How did you implement it:

`ts-node` dependency must be added by the user only if they use TypeScript.

## How can we verify it:

An error will be throw if the dependency is not added and the serverless-webpack configuration is written in TypeScript.

## Todos:

- [x] Write tests
- [ ] Write documentation
- [x] Fix linting errors
- [x] Make sure code coverage hasn't dropped
- [x] Provide verification config / commands / resources
- [x] Enable "Allow edits from maintainers" for this PR
- [x] Update the messages below

***Is this ready for review?:*** YES
***Is it a breaking change?:*** NO

Issue #633, opened 2020-09-01
serverless-webpack has unmet peer dependency


ZebraFlesh:
# This is a Bug Report

## Description
After doing an `npm install --save-dev serverless-webpack` I received a warning about unmet peer dependencies from npm:
```
npm WARN ts-node@8.10.2 requires a peer of typescript@>=2.7 but none is installed. You must install peer dependencies yourself.
```
My project doesn't use typescript, so this was unexpected. It seems like `ts-node` should be an [optionalDependency](https://docs.npmjs.com/files/package.json#optionaldependencies) instead of a dependency.

## Additional Data

* ***Serverless-Webpack Version you're using***: 5.3.4
* ***Webpack version you're using***: 4.44.1
* ***Serverless Framework Version you're using***: 1.80.0
* ***Operating System***: macOS 10.15.6 (Catalina)
* ***Stack Trace (if available)***: N/A

j0k3r:
I agree with you. But as I don't use TypeScript either I don't know how to properly fix it.

Does people using TypeScript always have the `ts-node` deps? How can we force them to install it? Because the deps is optional for people not using TypeScript but it's required if they use it.

How can we solve that?

Related PR https://github.com/serverless-heaven/serverless-webpack/pull/129

ZebraFlesh:
Based on [the instructions from webpack](https://v4.webpack.js.org/configuration/configuration-languages/#typescript), I'm pretty sure that people using TypeScript for their webpack file will have `ts-node` installed. That's why optional dependencies seems like the right answer. Looking at the [docs for `optionalDependencies`](https://docs.npmjs.com/files/package.json#optionaldependencies) npm notes:
> It is still your program’s responsibility to handle the lack of the dependency.

So perhaps `serverless-webpack` would need to provide a meaningful error message if the `require('ts-node')` call failed.

j0k3r:
Thanks for all these information!

j0k3r:
With 5.3.5 I now have:

```
warning "serverless-webpack > ts-node@9.0.0" has unmet peer dependency "typescript@>=2.7".
```

😕 
But there is nothing we can do now 🤷

PR #1825, opened 2024-06-15
fix: make `ts-node` dependency truly optional

G-Rath:
## What did you implement:

`ts-node` is meant to be an opt-in dependency (as in, a dependency that is not installed by default, but will be used if installed) per #636, but it's been added as an optional dependency rather than an .. err ... "right kind of optional dependency" (aka peer dependency set as optional).

In the context of `package.json`, "optional dependencies" are ones that [the package manager _should try to install_](https://docs.npmjs.com/cli/v10/configuring-npm/package-json#optionaldependencies) but skip if they error - they're intended for dependencies that are os-specific like `fsevents` or the underlying binaries for `esbuild` - so right now, `ts-node` will _always_ be installed because it does not have any os constraints, and in turn it leads to [complaints](https://github.com/serverless-heaven/serverless-webpack/issues/633) about [missing peer dependencies](https://github.com/serverless-heaven/serverless-webpack/pull/1287) via `ts-node`:

> If a dependency can be used, but you would like npm to proceed if it cannot be found or fails to install, then you may put it in the optionalDependencies object. This is a map of package name to version or URL, just like the dependencies object. The difference is that build failures do not cause installation to fail.

In the context of `package.json`, the correct way to express "optional" in the way most people think is to include the dependency as a peer dependency [with `optional: true` in its meta settings](https://docs.npmjs.com/cli/v10/configuring-npm/package-json#peerdependenciesmeta).

> specifically, it allows peer dependencies to be marked as optional. Npm will not automatically install optional peer dependencies. This allows you to integrate and interact with a variety of host packages without requiring all of them to be installed.

## How did you implement it:

I added `ts-node` as an optional peer dependency, and removed `typescript` and `@types/node` as peer dependencies since they're not used directly by `serverless-webpack`

## How can we verify it:

Install `serverless-webpack`, do an `npm ls` and see that `ts-node` isn't installed nor are there any dependency warnings - you can then install `typescript` and `ts-node`, and you'll still not have any dependency warnings.

Also see https://github.com/serverless/serverless-azure-functions/pull/517 for prior art

## Todos:

- [x] ~Write tests~
- [x] ~Write documentation~
- [x] ~Fix linting errors~
- [x] ~Make sure code coverage hasn't dropped~
- [x] ~Provide verification config / commands / resources~

***Is it a breaking change?:*** Debatable 😅

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -44,15 +44,11 @@
         "node": ">= 16"
       },
       "peerDependencies": {
-        "@types/node": "*",
         "serverless": "1 || 2 || 3",
         "ts-node": ">= 8.3.0",
         "webpack": ">= 3.0.0 < 6"
       },
       "peerDependenciesMeta": {
-        "@types/node": {
-          "optional": true
-        },
         "ts-node": {
           "optional": true
         }
--- a/package-lock.json
+++ b/package-lock.json
@@ -43,17 +43,18 @@
       "engines": {
         "node": ">= 16"
       },
-      "optionalDependencies": {
-        "ts-node": ">= 8.3.0"
-      },
       "peerDependencies": {
         "@types/node": "*",
         "serverless": "1 || 2 || 3",
+        "ts-node": ">= 8.3.0",
         "typescript": ">=2.0",
         "webpack": ">= 3.0.0 < 6"
       },
       "peerDependenciesMeta": {
         "@types/node": {
+          "optional": true
+        },
+        "ts-node": {
           "optional": true
         },
         "typescript": {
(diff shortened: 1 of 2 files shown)


j0k3r:
> _**Is it a breaking change?:**_ Debatable 😅

Well I do think people using TS with serverless-webpack already have `ts-node` as deps (at least we do). So moving it from a truly optional deps looks legit to me!

j0k3r (approved):

G-Rath:
> Well I do think people using TS with serverless-webpack already have ts-node as deps

Yup exactly my thoughts too

PR #1917, opened 2024-09-02
Apply promisify() to fs module functions (fixes #1912)

### LINKED ISSUE #1912: TypeError: fs.readFileAsync is not a function
# This is a Bug Report

## Description

When I run `serverless package`, I get error `fs.readFileAsync is not a function` and build fails:

```
⠹ Packaging (and 1 more task)^R

✖ fs.readFileAsync is not a function
TypeError: fs.readFileAsync is not a function
    at ServerlessWebpack.getFileContentAndStat (/PROJECT/node_modules/serverless-webpack/lib/packageModules.js:93:8)
    at /PROJECT/node_modules/serverless-webpack/lib/packageModules.js:57:74
    at arrayMap (/PROJECT/node_modules/lodash/lodash.js:653:23)
    at Function.map (/PROJECT/node_modules/lodash/lodash.js:9622:14)
    at WriteStream.<anonymous> (/PROJECT/node_modules/serverless-webpack/lib/packageModules.js:57:23)
    at WriteStream.emit (node:events:517:28)
    at WriteStream.emit (node:domain:489:12)
    at stream.emit (node:internal/fs/streams:61:9)
    at file:///Users/me/.serverless/releases/4.2.4/package/dist/sf-core.js:3:5724
    at file:///Users/me/.serverless/releases/4.2.4/package/dist/sf-core.js:3:6335
    at FSReqCallback.oncomplete (node:fs:200:23)
```

The error seems to come from this code block inside serverless-webpack:

```js
  return BbPromise.all([
    // Get file contents and stat in parallel
    fs.readFileAsync(fullPath),
    fs.statAsync(fullPath)
  ]).then(...)
```

As far as I could figure this out, the `fs.readFileAsync` is _not_ part of Node's fs module but that kind of (`*Async`) functions are created by "promisifying" a callback-based function. However, such promisifying is not done inside this package.

If I add `BbPromise.promisifyAll(fs);` below the imports in file `packageModules.js`, the build process succeeds.

I don't understand how this package works for other people if such promisifying step is missing.

## Additional Data

* ***Serverless-Webpack Version you're using***: 5.14.1
* ***Webpack version you're using***: 5.94.0
* ***Serverless Framework Version you're using***: 4.2.4
* ***Operating System***: OSX 14.6.1

ronkot:
## What did you implement:

Closes #1912

## How did you implement it:

Instead of relying on some other party applying promisify on `fs` module functions, do it explicitly inside this library. Also, apply promisify in non-mutable way to keep `fs` module tidy.

This should fix the `fs.readFileAsync is not a function` error when used with serverless v4.

## How can we verify it:

- Unit tests are OK
- Package some project with serverless v4 - process should not crash to the `fs.readFileAsync is not a function` error anymore
  - I tried to create serverless v4 example in the /examples folder but did not succeed. I think it's better if project maintainers would do that.

## Todos:

- [ ] Write tests
- [ ] Write documentation
- [ ] Fix linting errors

### DIFF
--- a/lib/packageModules.js
+++ b/lib/packageModules.js
@@ -9,6 +9,9 @@ const semver = require('semver');
 const fs = require('fs');
 const { getAllNodeFunctions, isProviderGoogle } = require('./utils');
 
+const readFileAsync = BbPromise.promisify(fs.readFile);
+const statAsync = BbPromise.promisify(fs.stat);
+
 function setArtifactPath(funcName, func, artifactPath) {
   const version = this.serverless.getVersion();
 
@@ -90,8 +93,8 @@ function getFileContentAndStat(directory, filePath) {
 
   return BbPromise.all([
     // Get file contents and stat in parallel
-    fs.readFileAsync(fullPath),
-    fs.statAsync(fullPath)
+    readFileAsync(fullPath),
+    statAsync(fullPath)
   ]).then(
     result => ({
       data: result[0],


j0k3r (approved):
Looks ok to me.

I tried to created a Serverless v4 example as well, but it requires a licence key, so I stopped here...

vicary (approved):
To put it lightly, v4 is not being easy on us plugin authors so far. Let's wait and see if we can actually cover these tests in the future.

The code looks fine and shouldn't do any harm at least, I'd say go ahead.

PR #1922, opened 2024-09-06
Update all examples

j0k3r:
Mostly:
- use latest version in `package.json`
- use `nodejs20.x` runtime
- add some missing README

I removed `serverless-offline` from plugins because it was never used


vicary (commented):
Since `serverless-offline` is being removed from `serverless.yml` in all examples, should we also remove them from `package.json` files?

vicary (approved):
LGTM

Issue #1923, opened 2024-09-06
Serverless 4 Support


jiteshmanglani:
# This is a Bug Report

## Description

For bug reports:
* Can't install packages due to unresolved dependency
* It should install all the packages
* normal config
* `npm ERR! code ERESOLVE
npm ERR! ERESOLVE could not resolve
npm ERR! 
npm ERR! While resolving: serverless-webpack@5.14.1
npm ERR! Found: serverless@4.2.4
npm ERR! node_modules/serverless
npm ERR!   dev serverless@"^4.2.4" from the root project
npm ERR!   peer serverless@">=2.60" from serverless-domain-manager@7.4.0
npm ERR!   node_modules/serverless-domain-manager
npm ERR!     dev serverless-domain-manager@"^7.4.0" from the root project
npm ERR! 
npm ERR! Could not resolve dependency:
npm ERR! peer serverless@"1 || 2 || 3" from serverless-webpack@5.14.1
npm ERR! node_modules/serverless-webpack
npm ERR!   dev serverless-webpack@"^5.14.1" from the root project
npm ERR! 
npm ERR! Conflicting peer dependency: serverless@3.39.0
npm ERR! node_modules/serverless
npm ERR!   peer serverless@"1 || 2 || 3" from serverless-webpack@5.14.1
npm ERR!   node_modules/serverless-webpack
npm ERR!     dev serverless-webpack@"^5.14.1" from the root project
npm ERR! 
npm ERR! Fix the upstream dependency conflict, or retry
npm ERR! this command with --force or --legacy-peer-deps
npm ERR! to accept an incorrect (and potentially broken) dependency resolution.
`

Similar or dependent issue(s):
* NA

## Additional Data

* ***Serverless-Webpack Version 5.14.1***:
* ***Webpack version 5.93.0***:
* ***Serverless Framework Version 4.2.4 ***:
* ***Operating System Fedora 39***:

j0k3r:
I checked quickly and it seems the v4 file structure completely changed.

v4 only expose the binary and that's all: https://www.npmjs.com/package/serverless?activeTab=code
v3 exported usefull stuff that we are using internally: https://www.npmjs.com/package/serverless/v/3.39.0?activeTab=code

Maybe it needs a better check to see if we can find a workaround.

vicary:
I've sent an email to Serverless, hopefully we get a reply soon.

vicary:
[...]

PR #1959, opened 2024-10-22
Add support for Serverless v4

j0k3r:
As of now, we can't jump straight to v4 in dev deps because some tests still require the v3 version (because we are mocking some stuff and the v4 only ship a binary).

The short solution is just to allow `4` in peer deps. But we prefer to have a running test that ensure a package is properly generated with deps inside.

As we can't use the internal tooling `@serverless/test` anymore in v4 (the project might be dead and no tests are running on the v4 official repo 😳, anyway) we are using the binary directly to package the example project. We were forced to use an access key from serverless dashboard and some aws valid credentials. But it now works ✅ 

Fix https://github.com/serverless-heaven/serverless-webpack/issues/1923


vicary (approved):
Totally counting on you for this one, thanks for making it happen.

PR #2409, opened 2026-04-21
Remove bluebird and use native Promise

j0k3r:
The code looks more native now.


vicary (approved):
Ah, finally!

Tried exactly this a few years ago, blocked by the support matrix back then. It surely feels weird that this is trivial work now with LLM.

PR #2479, opened 2026-07-08
Use `tinyglobby` instead of `glob`

j0k3r:
Way smaller than `glob`: https://e18e.dev/docs/replacements/glob

Once we'll be on Node >= 22 we might be able to ditch it for `fs.glob`

### DIFF
--- a/__mocks__/glob.js
+++ b/__mocks__/glob.js
@@ -1,7 +1,0 @@
-/**
- * Mock object for glob
- */
-
-module.exports = {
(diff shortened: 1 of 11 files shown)


vicary (dismissed):
I'm all for the change.

Partly OT but do we want to set a minimum published day for the recent surge of supply chain attacks?

j0k3r:
@vicary good point, I'll do it in an other PR 👍🏼

vicary (approved):

PR #2480, opened 2026-07-08
Define 7 days of minimum age for new deps

j0k3r:
Ensured in dependabot and in NPM

Following https://github.com/serverless-heaven/serverless-webpack/pull/2479#pullrequestreview-4653644899


vicary (approved):
lovely!

